PR #224, opened 2015-05-01
Add SSH banner as a metric in the lua ssh2 check

esproul:
We're already reading the banner; would be nice to have it as a metric.

### DIFF
--- a/src/modules-lua/noit/module/ssh2.lua
+++ b/src/modules-lua/noit/module/ssh2.lua
@@ -267,8 +267,8 @@ function initiate(module, check)
   local banner = e:read("\n")
   if banner == nil then return cerr("failed to read SSH banner") end
 
-  banner = banner:gsub("%s+$", "")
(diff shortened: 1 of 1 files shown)


src/modules-lua/noit/module/ssh2.lua
```diff
@@ -267,6 +267,9 @@ function initiate(module, check)
   local banner = e:read("\n")
   if banner == nil then return cerr("failed to read SSH banner") end
 
+  banner = banner:gsub("%s+$", "")
+  check.metric_string("banner", banner)
```
postwait: I wouldn't modify the banner variable.

drop the first line and just do:
```
check.metric_string("banner", banner:gsub("%s+$", ""))
```

esproul:
When I do that, I get a "wrong number of arguments" error.

esproul:
Using a local variable to get the trimmed version of the banner in.

PR #227, opened 2015-10-28
moved definition of histogram_t to histogram_impl.h

JonasKunze:
I moved the definition of histogram_t into histogram_impl.h so that the size is defined at compile time for external tools using this code.

### DIFF
--- a/src/modules/histogram_impl.c
+++ b/src/modules/histogram_impl.c
@@ -79,15 +79,6 @@ static double power_of_ten[256] = {
   1e-05, 0.0001, 0.001, 0.01, 0.1
 };
 
-struct histogram {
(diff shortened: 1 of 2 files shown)


postwait:
This breaks an intentional ABI safety.  This means that if we change the underlying implementation, people will need to recompile their apps.  That makes little sense.  The API itself allocated and destroys these histogram objects already, so there is no need for callers to know their size.  Why is this needed, this seems like a bad change.

JonasKunze:
I get your point and you're right. 
Without knowing the size you cannot create a histogram on the stack (and also not use std::unique_ptr). 
But it's okay for me to keep them in the heap instead, even though it will cost some performance.

PR #229, opened 2015-12-16
Added NoitMetricTransformed class

HeinrichHartmann:
This class carries an additional field "orderingId",
that can be used by other parts of the system.

= FOR REVIEW / DO NOT MERGE =

### DIFF
--- a/src/java/src/com/omniti/reconnoiter/event/NoitMetric.java
+++ b/src/java/src/com/omniti/reconnoiter/event/NoitMetric.java
@@ -60,6 +60,15 @@ public class NoitMetric extends NoitEvent
   }
   public NoitMetric(String[] parts) throws Exception {
     super(parts);
+
+    // Transform an MT-record back to a usual M-record for construction
+    if(parts[0].equals("MT")) {
+       String m_record[] = new String[7];
+       m_record[0] = "M";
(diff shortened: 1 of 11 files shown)


src/java/src/com/omniti/reconnoiter/event/NoitMetric.java
```diff
@@ -60,6 +60,15 @@ public NoitMetric(String[] parts, NoitMetricNumeric _nmn) throws Exception {
   }
   public NoitMetric(String[] parts) throws Exception {
     super(parts);
+
+    // Transform an MT-record back to a usual M-record for construction
+    if(parts[0].equals("MT")) {
+       String m_record[] = new String[7];
+       m_record[0] = "M";
+       for(int i=1;i<7;i++) { m_record[i] = parts[i]; }
```
JonasKunze: Is somewhere checked that parts.length is actually 7?
If now we'd also have the same problem at line 72: parts[5]

postwait: Yes. super -> super -> https://github.com/HeinrichHartmann/reconnoiter/blob/master/src/java/src/com/omniti/reconnoiter/StratconMessage.java#L123-L125

HeinrichHartmann:
No longer needed.

PR #236, opened 2016-04-20
fix warnings

rileyberton:
Compile warnings under -Wall -Werror have been fixed.. mostly this is unused variables.

### DIFF
--- a/src/modules/histogram.c
+++ b/src/modules/histogram.c
@@ -28,7 +28,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <sys/int_fmtio.h>
+#include <inttypes.h>
 #include <mtev_defines.h>
 
 #include <assert.h>
--- a/src/modules/histogram.c
+++ b/src/modules/histogram.c
@@ -28,6 +28,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <sys/int_fmtio.h>
 #include <mtev_defines.h>
 
 #include <assert.h>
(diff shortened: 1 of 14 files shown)


src/modules/histogram.c
```diff
@@ -28,6 +28,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <sys/int_fmtio.h>
```
aidancully: shouldn't this be `#include <inttypes.h>`?

src/modules/lua_check.c
```diff
@@ -698,11 +696,10 @@ noit_lua_module_config(noit_module_t *mod,
 
 static int
 noit_lua_module_init(noit_module_t *mod) {
-  struct module_conf *mc;
   struct module_tls_conf *mtlsc;
   LMC_DECL(L, mod, object);
 
-  mc = noit_module_get_userdata(mod);
+  noit_module_get_userdata(mod);
```
aidancully: I think this is a no-op? just remove the call?

src/noit_conf_checks.c
```diff
@@ -41,6 +41,7 @@
 #include <libxml/parser.h>
 #include <libxml/tree.h>
 #include <libxml/xpath.h>
+#include <sys/int_fmtio.h>
```
aidancully: probably should be `inttypes.h`

rileyberton: Ahh yes.  Good point.

src/modules/lua_check.c
rileyberton: Not sure what side-effects get_userdata has on `mod`

aidancully: If I read [this](https://github.com/circonus-labs/reconnoiter/blob/master/src/noit_module.c#L326-L334) right, none.

rileyberton: Right, will remove.

rileyberton:
merged by hand

PR #243, opened 2016-05-18
initial implementation of a metrics observer

postwait:


src/noit_metric_director.c
```diff
@@ -0,0 +1,231 @@
...
+  ck_fifo_spsc_dequeue_unlock(my_lane.fifo);
+  return line;
+}
+static void
+handle_metric_buffer(const char *payload, int payload_len, int has_noit) {
+  switch(payload[0]) {
+    case 'H':
+    case 'M':
+      {
+        struct metric metric;
```
igalic: this reads super confusing…

postwait: Is the confusion specific to the variable being named the same thing as the struct?

igalic: yes

postwait: Fixed.  and refactored a bit.

PR #246, opened 2016-05-25
implemented noit_message_decoder

JonasKunze:


JonasKunze:
Will reopen as soon as I'm done with the missing metatables

src/modules/noit_lua_noit_binding.c
```diff
@@ -126,12 +142,210 @@ lua_noit_checks_unsubscribe(lua_State *L) {
...
+  udata = lua_touserdata(L, 1);
+  msg = *udata;
+  if(!lua_isstring(L, 2)) {
+    luaL_error(L, "metatable error, arg2 not a string!");
+  }
+
+  k = lua_tostring(L, 2);
+  switch(*k) {
+    case 'i':
+      noit_lua_setup_metric_id(L, &msg->id);
```
postwait: this will allow sloppy usage like metric.id:isndifunsdf

you should

```
if(!strcmp(k, "id")) {
  noit_lua_setup_metric_id(L, &msg->id);
  return 1;
}
break;
```

And on all the subsequent ones as well.

src/noit_message_decoder.c
```diff
@@ -0,0 +1,165 @@
...
+    if(vstrlen == 0)
+      metric->value.v_string = NULL;
+    else {
+      metric->value.v_string = value_str;
+    }
+    return 1;
+  }
+
+  if(*payload == 'S' || *payload == 'C' || *payload == 'D') {
+    metric->type = METRIC_STRING;
```
postwait: The S, C and D records aren't metrics at all... this is a sloppy abstraction.

the problem here is that the directory is really a message director I think trying to squeeze non-metric types back into a strongly typed metric is a bad idea.

postwait: How about setting this to METRIC_NONE and setting the v_string to NULL;

src/modules/noit_lua_noit_binding.c
```diff
@@ -126,12 +142,227 @@ lua_noit_checks_unsubscribe(lua_State *L) {
...
+  }
+  udata = lua_touserdata(L, 1);
+  metric_id = *udata;
+  if(!lua_isstring(L, 2)) {
+    luaL_error(L, "metatable error, arg2 not a string!");
+  }
+
+  switch (*k) {
+    case 'w':
+      k = lua_tostring(L, 2);
```
postwait: This should go a few lines up, no? above the switch?

src/noit_message_decoder.c
```diff
@@ -0,0 +1,165 @@
...
+ * modification, are permitted provided that the following conditions are
+ * met:
+ * 
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above
+ *       copyright notice, this list of conditions and the following
+ *       disclaimer in the documentation and/or other materials provided
+ *       with the distribution.
+ *     * Neither the name OmniTI Computer Consulting, Inc. nor the names
```
postwait: This should say Circonus, Inc. (not OmniTI Computer Consulting, Inc.)

src/noit_message_decoder.h
```diff
@@ -0,0 +1,59 @@
...
+ * modification, are permitted provided that the following conditions are
+ * met:
+ * 
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above
+ *       copyright notice, this list of conditions and the following
+ *       disclaimer in the documentation and/or other materials provided
+ *       with the distribution.
+ *     * Neither the name OmniTI Computer Consulting, Inc. nor the names
```
postwait: This should say Circonus, Inc. (not OmniTI Computer Consulting, Inc.)

src/modules/noit_lua_noit_binding.c
```diff
@@ -69,6 +69,22 @@ nl_check(lua_State *L) {
 }
 
 static int
+nl_check_ud(lua_State *L) {
+  uuid_t *uuid;
+  uuid = lua_touserdata(L, 1);
+
+  noit_check_t *check;
+
+  check = noit_poller_lookup(*uuid);
```
rileyberton: If we're not going to be strictly c89 compliant anyway, you can declare the check and init in the same line.. same with uuid above.

src/modules/noit_lua_noit_binding.c
JonasKunze: oh yes! that's why I'm currently focusing on tests...

src/modules/noit_lua_noit_binding.c
```diff
@@ -126,12 +142,227 @@ lua_noit_checks_unsubscribe(lua_State *L) {
...
+  if(!luaL_checkudata(L, 1, "metric_value_t")) {
+    luaL_error(L, "metatable error, arg1 not a metric_value_t!");
+  }
+  udata = lua_touserdata(L, 1);
+  metric_id = *udata;
+  if(!lua_isstring(L, 2)) {
+    luaL_error(L, "metatable error, arg2 not a string!");
+  }
+
+  switch (*k) {
```
rileyberton: Derefing uninited?

src/modules/noit_lua_noit_binding.c
```diff
@@ -126,12 +142,227 @@ lua_noit_checks_unsubscribe(lua_State *L) {
...
+  metric_id = *udata;
+  if(!lua_isstring(L, 2)) {
+    luaL_error(L, "metatable error, arg2 not a string!");
+  }
+
+  k = lua_tostring(L, 2);
+  switch (*k) {
+    case 'i':
+      if(!strcmp(k, "id")) {
+        char uuid_str[UUID_STR_LEN + 1];
```
rileyberton: "string of the length defined in the uuid.h macro,
       UUID_PRINTABLE_STRING_LENGTH, which includes the trailing null
       character."   

Could use `UUID_PRINTABLE_STRING_LENGTH` instead of the +1

src/modules/noit_lua_noit_binding.c
```diff
@@ -126,12 +142,227 @@ lua_noit_checks_unsubscribe(lua_State *L) {
...
+  if(!luaL_checkudata(L, 1, "metric_message_t")) {
+    luaL_error(L, "metatable error, arg1 not a metric_message_t!");
+  }
+  udata = lua_touserdata(L, 1);
+  msg = *udata;
+  if(!lua_isstring(L, 2)) {
+    luaL_error(L, "metatable error, arg2 not a string!");
+  }
+
+  k = lua_tostring(L, 2);
```
rileyberton: Is it possible this returns NULL?  Will crash on next line if so.

src/noit_message_decoder.c
```diff
@@ -0,0 +1,165 @@
...
+    cp = lvalue; \
+    } \
+} while(0)
+
+int noit_message_decoder_parse_M_OR_H_line(const char *payload, int payload_len,
+    uuid_t *id, const char **metric_name, int *metric_name_len,
+    metric_value_t *metric, mtev_boolean has_noit) {
+  const char *cp, *metric_type_str, *time_str, *check_id_str;
+  char *value_str;
+  char *dp, id_str_copy[UUID_STR_LEN + 1];
```
rileyberton: `UUID_PRINTABLE_STRING_LENGTH`

src/noit_metric_director.c
```diff
@@ -154,76 +131,100 @@ noit_adjust_metric_interest(uuid_t id, const char *metric, short cnt) {
...
-  for(i=0;i<nthreads;i++) {
+  for(i = 0; i < nthreads; i++) {
     if(interests[i] > 0) {
-      char *copy = mtev__strndup(line, line_len);
-      ck_fifo_spsc_t *fifo = (ck_fifo_spsc_t *)thread_queues[i];
+      ck_fifo_spsc_t *fifo = (ck_fifo_spsc_t *) thread_queues[i];
       ck_fifo_spsc_entry_t *fifo_entry;
       fifo_entry = ck_fifo_spsc_recycle(fifo);
       if(!fifo_entry) fifo_entry = malloc(sizeof(ck_fifo_spsc_entry_t));
       ck_fifo_spsc_enqueue_lock(fifo);
```
rileyberton: This lock needs to move above the `recycle` line.  `recycle` is not MP safe so the lock should protect it.

src/noit_metric_director.c
```diff
@@ -154,76 +131,100 @@ noit_adjust_metric_interest(uuid_t id, const char *metric, short cnt) {
...
 }
 static void
-handle_metric_buffer(const char *payload, int payload_len, int has_noit) {
-  switch(payload[0]) {
+handle_metric_buffer(const char *payload, int payload_len,
+    int has_noit) {
+  // mtev_fq will free the fq_msg -> copy the payload
+  char *copy = mtev__strndup(payload, payload_len);
+
+  metric_message_t *message = calloc(1, sizeof(metric_message_t));
```
rileyberton: Who frees `message`?

src/modules/noit_lua_noit_binding.c
postwait: Is that available in all our uuid.h's on all platforms.  That UUID_STR_LEN+1 is my pattern b/c I couldn't find anything in the includes when I looked.  If it is "the right way", we should search and replace the whole damn tree -- because that is everywhere.

src/noit_metric_director.c
```diff
@@ -154,76 +131,100 @@ noit_adjust_metric_interest(uuid_t id, const char *metric, short cnt) {
...
+  if(ck_fifo_spsc_dequeue(my_lane.fifo, &msg) == false) {
+    msg = NULL;
   }
   ck_fifo_spsc_dequeue_unlock(my_lane.fifo);
-  return line;
+  return msg;
+}
+void noit_metric_director_free_message(metric_message_t* message) {
+  if(message->original_message) {
+    free(message->original_message);
```
rileyberton: So this frees the original string but who frees the calloc here on line 201?

src/modules/noit_lua_noit_binding.c
JonasKunze: UUID_STR_LEN is defined in `mtev_defines`. There's no `UUID_PRINTABLE_STRING_LENGTH`! I can add this!

src/modules/noit_lua_noit_binding.c
JonasKunze: I'm returning now if the 2nd parameter is not a string

src/noit_metric_director.c
JonasKunze: fixed (see line 194)

src/noit_metric_director.c
JonasKunze: noit_metric_director_free_message does now!

rileyberton:
Looks good to me

PR #252, opened 2016-06-07
Fix detection of libmtev in configure.in

flyerman:


postwait:
Shouldn't libmtev depend on -lck and -ljlog already?  An app should be able to -lmtev and have it automatically pull in all dependent libs (include ck and jlog).

On what platform is this error happening?

flyerman:
It does indeed. The issue is when installing libck and libjlob in custom locations.
This happening on CentOS.

postwait:
Did you build libmtev with -Wl,-rpath -Wl,/custom/lib/location ?

It sounds like libmtev was build without a correct runpath

flyerman:
I have not used rpath. I'll take a look thanks.

PR #254, opened 2016-06-08
Fix noit connection

rileyberton:


src/js/noit-connection/noit-connection.js
```diff
@@ -234,7 +234,15 @@ function handle_frame(parent, frame, host, port) {
...
-              parent.socket.write(frame_output(chan.id, false, buff));
+              var bl = buff.length;
+              var os = 0;
+              while (os < bl) {
+                  var tb_size = Math.min(MAX_FRAME_LEN, bl - os);
+                  var tempbuff = new Buffer(tb_size);
+                  buff.copy(tempbuff, 0, os, os + tb_size);
+                  parent.socket.write(frame_output(chan.id, false, tempbuff));
+                  os += tempbuff.length;
+              }
```
postwait: If the buff.length is already <= MAX_FRAME_LEN, how about simple write that avoid the buff.copy?

postwait:
+1

PR #256, opened 2016-06-09
Correct functions into the mtev namespace that belong there.

postwait:


rileyberton:
seems fairly mechanical.  LGTM

PR #258, opened 2016-06-15
Force Check Lookup In HttpTrap Callback

pamaddox:
Every time we get into the HttpTrap callback, force looking up the check
by the uuid. This will prevent us from trying to use a stale check
pointer.

### DIFF
--- a/src/modules/httptrap.c
+++ b/src/modules/httptrap.c
@@ -80,6 +80,7 @@ struct value_list {
 };
 struct rest_json_payload {
   noit_check_t *check;
+  uuid_t check_id;
   yajl_handle parser;
   int len;
   int complete;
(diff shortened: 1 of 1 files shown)


src/modules/httptrap.c
```diff
@@ -457,6 +458,12 @@ rest_get_json_upload(mtev_http_rest_closure_t *restc,
 
   content_length = mtev_http_request_content_length(req);
   rxc = restc->call_closure;
+  rxc->check = noit_poller_lookup(rxc->check_id);
+  if (!rxc->check) {
+    *complete = 1;
```
aidancully: should we report an error, here?

aidancully: should we report an error, here?

pamaddox: This isn't necessarily an error.... it likely means that a check was deleted while in the middle of processing and that the rest of the data should be discarded. I think we could potentially spam too much stuff in the error log here if people constantly create and delete a bunch of httptrap checks.

aidancully: I meant to clients... I guess an ```
rxc->error = strdup("check invalidated during processing");
``` or something like that?

aidancully: oic, we report "internal error" in that case, nvm.

aidancully:
LGTM

PR #266, opened 2016-06-23
Use mtevFatal Instead Of assert/abort

pamaddox:
Replaces tons of assert/abort calls with calls to mtevFatal, which logs
synchronously before exiting. This is better for debugging - it makes
sure that we get errors in logs before we abort/assert out.

I didn't replace every single instance, since there are tons of them,
but I replaced a bunch of the more important ones.

### DIFF
--- a/src/modules/collectd.c
+++ b/src/modules/collectd.c
@@ -27,7 +27,6 @@
 #include <stdio.h>
 #include <unistd.h>
 #include <errno.h>
-#include <assert.h>
 #include <math.h>
 #include <ctype.h>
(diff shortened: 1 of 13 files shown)


aidancully:
The patch looks correct, though I'd prefer to define / use a `mtevAssert` macro, instead? Then we wouldn't need to invert all the conditions, and covering all the `assert` / `abort` can conceivably be done with a simple sed script.

pamaddox:
I think it'd be better to write a macro for asserts rather than changing everything to use the mtevFatal stuff when we need to do asserts.... gonna close this pull request.

PR #267, opened 2016-06-23
Make Lua SNMP Loading Safer

pamaddox:
Two changes:

* Only load lua_snmp if there's a jezebel snmp lua module defined
* Lock around the lua_snmp initialization to make sure that multiple
  threads don't call it at once, make sure that we only init once


postwait:
LGTM

PR #268, opened 2016-06-23
Use mtevAssert and mtevFatal Instead Of assert() and abort()

pamaddox:
Use libmtev calls to safely flush logs and abort rather than calling
the assert and abort calls directly.


aidancully:
LGTM

PR #270, opened 2016-06-27
fix `make_conf_path`'s null-termination of local `fullpath` variable.

aidancully:
was using `strlen(fullpath)` where `sizeof(fullpath)` was intended.
fixed.


rileyberton:
Looks good to me

PR #271, opened 2016-06-27
Fix Memory Leaks In noit_metric_director.c

pamaddox:
The noit metric directory code was very leaky.... fixed the memory leaks
in it.


pamaddox:
Going in a different direction with this.

PR #273, opened 2016-06-27
Fix So We Don't Have A Use-After-Free On Passive Checks

pamaddox:
If a check is passive, recycle it unless it's either:

* Flagged as running for some reason (shouldn't happen)
or....
* Last ran (period*2) seconds ago

This will prevent freeing a check that's in-flight and causing possible
memory errors.

### DIFF
--- a/src/noit_check.c
+++ b/src/noit_check.c
@@ -1325,7 +1325,7 @@ noit_poller_free_check(noit_check_t *checker) {
     struct timeval current_time;
     gettimeofday(&current_time, NULL);
     if (checker->last_fire_time.tv_sec == 0) {
-      gettimeofday(&checker->last_fire_time, NULL);
+      memcpy(&checker->last_fire_time, &current_time, sizeof(struct timeval));
     }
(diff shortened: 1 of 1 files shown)


src/noit_check.c
```diff
@@ -1321,7 +1321,18 @@ void
...
   noit_module_t *mod;
 
-  if(checker->flags & NP_RUNNING) {
+  if (checker->flags & NP_PASSIVE_COLLECTION) {
+    struct timeval current_time;
+    gettimeofday(&current_time, NULL);
+    /* If NP_RUNNING is set for some reason or we've fired recently, recycle
+     * the check.... we don't want to free it */
+    if ((checker->flags & NP_RUNNING) ||
+        ((current_time.tv_sec - checker->last_fire_time.tv_sec) < (checker->period*2))) {
```
postwait: tv_sec is in seconds, period is in milliseconds.

aidancully: use `sub_timeval` (or `sub_timeval_d`) to calculate tv difference?

postwait:
LGTM

PR #274, opened 2016-06-27
make check recycling thread safe

postwait:


pamaddox:
Looks fine to me.

src/noit_check.c
```diff
@@ -1331,12 +1334,12 @@ noit_poller_free_check(noit_check_t *checker) {
      * the check.... we don't want to free it */
     if ((checker->flags & NP_RUNNING) ||
         (sub_timeval_ms(current_time,checker->last_fire_time) < (checker->period*2))) {
-      recycle_check(checker);
+      recycle_check(checker, has_lock);
```
pamaddox: I think we want to return after this.

pamaddox:
I added some fixes here:

https://github.com/circonus-labs/reconnoiter/pull/276

Think we want to revisit this after we merge PR 276.

pamaddox:
276 has been merged, but it appears to have caused some conflicts here that need addressing.

pamaddox:
Tested a bunch, seems to work fine. Looks good.

PR #277, opened 2016-06-28
Free Histotier Object When Done With It

pamaddox:
We were freeing the contents of the histotier object, but not the object
itself. Free it.


rileyberton:
Looks good to me

PR #279, opened 2016-07-01
supporting [[null]] message values

JonasKunze:


src/noit_message_decoder.c
```diff
@@ -98,37 +98,33 @@ int noit_message_decoder_parse_line(const char *payload, int payload_len,
...
-      metric->value.v_double = strtod(value_str, NULL);
-      break;
-    case METRIC_STRING:
-      metric->type = METRIC_STRING;
-      if(!strcmp(value_str, "[[null]]"))
-        metric->value.v_string = NULL;
-      else {
+    metric->type = *metric_type_str;
+    if(!strcmp(value_str, "[[null]]")) {
+      metric->type = METRIC_ABSENT;
```
postwait: Most of the rest of the system still wants the type information. You want to know if itis an absent string or double or int.

JonasKunze:
Then I'd suggest to add a field "is_absent" and leave the type as is because currently there's no way to store both information `absent` and `type` at once

JonasKunze:
More efficient would be to store the original type in `value` and set the type to `ABSENT` -> no additional field required

src/noit_message_decoder.c
postwait: So. This isnt a metric_t? B/c that, by design, has all elements as pointers rhat can be nulled. Maybe we should change the datatype here.

JonasKunze: no it isn't!

```
typedef struct {
  uint64_t whence_ms; /* when this was recieved */
  metric_type_t type; /* the type of the following data item */
  union {
    int32_t v_int32;
    uint32_t v_uint32;
    int64_t v_int64;
    uint64_t v_uint64;
    double v_double;
    char *v_string;
  } value; /* the data itself */
} noit_metric_value_t;
```

JonasKunze: It's not quite the same as metric_c:
```
typedef struct {
  char *metric_name;
  metric_type_t metric_type;
  union {
    double *n;
    int32_t *i;
    u_int32_t *I;
    int64_t *l;
    u_int64_t *L;
    char *s;
    void *vp; /* used for clever assignments */
  } metric_value;
  mtev_boolean logged;
} metric_t;
```
I think we should stay with two types!

postwait: Okay.. well the problem with the other type is that it can't express a NULL METRIC_DOUBLE... can we fix that so it can?

JonasKunze: see last commit

postwait:
LGTM

PR #281, opened 2016-07-05
Make httptrap truncate metric names to 255 and continue.

postwait:
Note truncation errors in the successful response.
Use mtev_json to correctly output JSON responses.


rileyberton:
Looks good to me

PR #285, opened 2016-07-12
Fixed undefined symbol issue and cleared compiler warning

JonasKunze:


pamaddox:
This looks fine to me.

JonasKunze:
There seems to be another issue with that:
Cannot find 'lua' in image '/opt/noit/prod/libexec/noit/lua.so': ld.so.1: noitd: fatal: lua: can't find symbol

PR #286, opened 2016-07-13
Fixed undefined symbol issue and cleared compiler warning

JonasKunze:
I'm not sure if it's a good idea to let lua.so AND noit_lua/noit_binding.so depend on $(LUA_MODULE_OBJS) but this at least fixed my issues with undefined symbols in either lua.so or noit_binding.so


pamaddox:
This seems fine.

PR #287, opened 2016-07-13
moved noit_metric_director_init() in front of the module initialization

JonasKunze:
to be able to use e.g. `noit.checks_do` in lua `noit_metric_director_init()` must have been called first


pamaddox:
Looks fine to me.

PR #288, opened 2016-07-14
Call noit_mtev_bridge_init After libmtev Init

pamaddox:
noit_mtev_bridge_init uses hash tables initialized in libmtev's init
function.... we need to make sure we call this AFTER we initialize
those, not before.


aidancully:
LGTM

PR #291, opened 2016-07-19
initializing lmc->pending

JonasKunze:
without that I godd flooded with `warning: null hashtable in mtev_hash_store... initializing`


pamaddox:
Looks good to me.

PR #293, opened 2016-07-21
Lock Hashes In noit_check.c

pamaddox:
There are a bunch of hashes in noit_check.c that could theoretically get
hit from multiple threads.... initialize them as locking.


postwait:
The assumption is correct.  This is needed.  LGTM.

PR #297, opened 2016-07-28
Reverted a8a86eb: mtev_metric_director doesn't connect to fq otherwise

JonasKunze:


postwait:
LGTM. Startup order can be hairy

PR #299, opened 2016-08-05
BC and FC key naming fix

postwait:


pamaddox:
If it's assumed that the partial matching is good/intended, then this should be good.

aidancully:
The documentation for reverse_check still says "key", should probably be updated to "secret_key"?

aidancully:
LGTM

PR #300, opened 2016-08-05
Support { "_fl": "+|~" } operations to accumulate and average.

postwait:
If httptrap submissions contain an extended value with _fl, it
will attempt to accumulate (+) continuously or average (~) over
the period.  If the logging is immediate, averaging over the
period is avoided for replacement as individually reported
values will be averaged upstream.


maier:
nice, this seems like a workable solution

PR #304, opened 2016-08-18
Bugfix: rollup has to access last value even if it's in the last window

JonasKunze:
The derivative is calculated based on the last value seen, even if it's not in the current window. Hence, even if the current window is empty (accu->accumulated.type == METRIC_NULL) the derivative can be calculated.


postwait:
LGTM

PR #305, opened 2016-08-24
noit_stats_log_immediate_metric didn't add the metric to the global metric table

JonasKunze:
hence we couldn't see the metrics dispatched via noit_stats_log_immediate_metric in /checks/show/account_<X>/<UUID>

### DIFF
--- a/src/noit_check.c
+++ b/src/noit_check.c
@@ -1731,12 +1731,18 @@ __stats_add_metric(stats_t *newstate, metric_t *m) {
                     m, NULL, (void (*)(void *))mtev_memory_safe_free);
 }
 
-static void
-__mark_metric_logged(stats_t *newstate, const char *metric_name) {
+static mtev_boolean
+__mark_metric_logged(stats_t *newstate, metric_t *m) {
   void *vm;
   if(mtev_hash_retrieve(&newstate->metrics,
-                        metric_name, strlen(metric_name), &vm)) {
+      m->metric_name, strlen(m->metric_name), &vm)) {
     ((metric_t *)vm)->logged = mtev_true;
+    return mtev_false;
+  } else {
+    m->logged = mtev_true;
+    mtev_hash_replace(&newstate->metrics, m->metric_name, strlen(m->metric_name),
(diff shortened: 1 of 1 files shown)


src/noit_check.c
```diff
@@ -1732,11 +1732,14 @@ __stats_add_metric(stats_t *newstate, metric_t *m) {
 }
 
 static void
-__mark_metric_logged(stats_t *newstate, const char *metric_name) {
+__stats_add_and_logg_metric(stats_t *newstate, metric_t *m) {
```
postwait: I think you want to leave the function named __mark_mertric_logged.

src/noit_check.c
```diff
@@ -1732,11 +1732,14 @@ __stats_add_metric(stats_t *newstate, metric_t *m) {
...
   void *vm;
   if(mtev_hash_retrieve(&newstate->metrics,
-                        metric_name, strlen(metric_name), &vm)) {
+      m->metric_name, strlen(m->metric_name), &vm)) {
     ((metric_t *)vm)->logged = mtev_true;
+  } else {
+    mtev_hash_replace(&newstate->metrics, m->metric_name, strlen(m->metric_name),
+                        m, NULL, (void (*)(void *))mtev_memory_safe_free);
   }
```
postwait: after this, m->logged isn't set to true, so you need to set that I think.

src/noit_check.c
```diff
@@ -1732,11 +1732,14 @@ __stats_add_metric(stats_t *newstate, metric_t *m) {
 }
 
 static void
-__mark_metric_logged(stats_t *newstate, const char *metric_name) {
+__stats_add_and_logg_metric(stats_t *newstate, metric_t *m) {
   void *vm;
   if(mtev_hash_retrieve(&newstate->metrics,
-                        metric_name, strlen(metric_name), &vm)) {
+      m->metric_name, strlen(m->metric_name), &vm)) {
```
postwait: You need to mtev_memory_safe_free(m) here if m != vm (which it likely does not)

postwait:
LGTM

PR #306, opened 2016-08-25
Immediate histo

JonasKunze:
This allows us to publish histograms from lua via `check.immediate_histogram(metric, encoded_histo, whence_s)`

The lua part is already tested by the caql broker, the refactored version of `histogram.c` is not however.

### DIFF
--- a/src/modules/histogram.c
+++ b/src/modules/histogram.c
@@ -29,7 +29,6 @@
  */
 
 #include <inttypes.h>
-#include <mtev_defines.h>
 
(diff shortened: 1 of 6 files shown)


postwait:
This  looks good to me in concept, I assume you've tried it out.

JonasKunze:
Yes, but I'd like to do some more tests before merging this.

JonasKunze:
may I merge this?

postwait:
Lgtm

PR #308, opened 2016-08-29
mtev_log_line_hook_register takes CONST timeval

JonasKunze:


postwait:
LGTM contingent upon the libmtev PR that supports this.

PR #309, opened 2016-08-30
Timed immediate metric

JonasKunze:
This allows the following lua code:
check.immediate_metric_double('name', 1.0, 123, 456)

This value should then be distributed at the timestamp 123.456
NOTE: currently for some reason the usecs are ignored. So the actual M message stores the time 123.000

### DIFF
--- a/src/noit_check.c
+++ b/src/noit_check.c
@@ -151,7 +151,7 @@ struct stats_t {
 };
 
 struct timeval *
-noit_check_stats_whence(stats_t *s, struct timeval *n) {
+noit_check_stats_whence(stats_t *s, const struct timeval *n) {
   if(n) memcpy(&s->whence, n, sizeof(*n));
   return &s->whence;
 }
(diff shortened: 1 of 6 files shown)


src/noit_check.c
```diff
@@ -1967,12 +1967,23 @@ noit_stats_set_metric(noit_check_t *check,
...
+}
+void
+noit_stats_set_metric_timed(noit_check_t *check,
+                      const char *name, metric_type_t type,
+                      const void *value, const struct timeval *time) {
+  stats_t *c;
+  c = noit_stats_set_metric(check, name, type, value);
+  if(c) {
+    noit_check_stats_whence(c, time);
+  }
```
postwait: This function sets the time for the whole stats structure, not just the metric.  I'm not sure we want to support such operations.  What's the expected use case where a caller will set the time for a "metric" but be fine with all metrics in the stats structure having their container's timestamp set?

postwait:
LGTM

PR #311, opened 2016-09-16
Fix Crash When NULL Item Found In Array

pamaddox:
We were crashing on a null JSON array object.... check for the
condition, log, and move on when we hit it.


rileyberton:
Looks good to me

PR #313, opened 2016-09-21
MongoDB Check

neophenix:
The branch name is a lie, instead of trying to coerce the mongo client stuff to JDBC I've just made it its own check.  Maybe I need to revert the changes to the JDBC files a different way?  Dunno


postwait:
I don't see any docs for it. I assume that is b/c it masquerades as the jezebel check?  Not sure what the right solution for that is.

neophenix:
Yeah, none of the jezebel ones do that I know of.  I have the "typical" XML for it, since I needed that to integrate this check, I guess I could just plop that into mongodb.java.  Would still be missing it for all the others.

postwait:
Actually.. I think dropping it in /src/modules/mongodb.xml might do the trick.

Then we can think about better organization later.

src/modules/mongodb.xml
```diff
@@ -0,0 +1,50 @@
...
+               allowed=".+">The command that will be executed</parameter>
+  </checkconfig>
+  <examples>
+    <example>
+      <title>MongoDB Server Status</title>
+      <para>This will run the serverStatus command every 5 minutes against the mongodb server at 10.225.209.36 with the database mydb logging in as user / pass</para>
+      <programlisting><![CDATA[
+      <noit>
+        <modules>
+          <module image="mongodb" name="mongodb"/>
```
postwait: That doesn't look right.
```
<module>
  <loader image="lua" name="lua"/>
  <module loader="lua" name="mongodb" object="noit.module.jezebel"/>
</modules>
```

postwait:
LGTM

PR #314, opened 2016-09-28
Implemented noit_check_path and wired it to check.path in lua

JonasKunze:


src/modules/lua_check.c
```diff
@@ -695,6 +695,12 @@ noit_check_index_func(lua_State *L) {
       return 1;
     case 'p':
       if(!strcmp(k, "period")) lua_pushinteger(L, check->period);
+      else if(!strcmp(k, "path")) {
+        int rv = 0;
```
rileyberton: what's `rv` for?

JonasKunze: just a relict..sorry for this!

rileyberton:
Looks good to me

PR #315, opened 2016-09-30
mark ping_icmp noit-module as thread-unsafe.

aidancully:
It only uses two FD sockets for ICMP responses, so it is mostly
single-threaded already. There is no benefit to having the
noit check infrastructure itself in another thread, and some risk
to doing so. So, make the ping_icmp's noit_module single-threaded.


pamaddox:
Seems reasonable to me.

PR #316, opened 2016-10-04
Bugfix: locking when calling mtev_conf in noit_check

JonasKunze:


src/noit_check.c
```diff
@@ -69,6 +69,8 @@
 #define CHECKS_XPATH_PARENT "checks"
 #define CHECKS_XPATH_BASE CHECKS_XPATH_ROOT "/" CHECKS_XPATH_PARENT
 
+static pthread_mutex_t conf_lock;
```
aidancully: `static pthread_mutex_t conf_lock = PTHREAD_MUTEX_INITIALIZER;`

postwait:
I have no idea how this would fix a problem.  The only way this could be causing an issue was if someone was *writing* to the XML at the same time and this patch doesn't lock around modifications to the XML.

JonasKunze:
Damn, I saw your comment only now, after merging...

That was actually my question yesterday. I also thought that `mtev_conf_get_section` is a read only function and hence should be thread safe. But without that log only about 10 out of 2000 times the returned node is not NULL. With that lock the returned node is never NULL again.
I'll try to find out who's writing to the XML and understand why this fixes my issue

PR #318, opened 2016-10-18
Storing path of every check to bypass unsafe mtev_con access from lua…

JonasKunze:
… threads


JonasKunze:
As Theo mentioned the data I'm looking (account_id) for is also stored in the check name -> don't need this any more

PR #319, opened 2016-10-18
Have cloudwatch treat missing data like all other checks

neophenix:
Currently, cloudwatch pushes missing metrics into the response with
a null value.  This is unlike other checks that omit them entirely
and causes confusion further down the line.

We should just treat them line any other check would, and not put
them in the response


pamaddox:
Should be fine

PR #320, opened 2016-10-19
removed non thread safe check.path from lua

JonasKunze:
This reverts some parts of https://github.com/JonasKunze/reconnoiter/commit/97c5c2142634286fde3578e79cc2af284c45ee38
I've left some parts of this commit (essentially noit_check_path) as I'm using it in another feature branch


postwait:
LGTM

PR #324, opened 2016-10-25
added optional seq to filtersets

JonasKunze:
This allows the UI to push rulesets with a seq tag in it. We'll need these for a clustered noit soon.

### DIFF
--- a/src/noit_filters.c
+++ b/src/noit_filters.c
@@ -86,7 +86,7 @@ typedef struct _filterrule {
 typedef struct {
   mtev_atomic32_t ref_cnt;
   char *name;
-  int64_t seq;
+  uint64_t seq;
   filterrule_t *rules;
 } filterset_t;
 
@@ -130,6 +130,7 @@ noit_filter_compile_add(mtev_conf_section_t setinfo) {
   int j, fcnt;
   char filterset_name[256];
   filterset_t *set;
+  int64_t seq = 0;
   if(!mtev_conf_get_stringbuf(setinfo, "@name",
                               filterset_name, sizeof(filterset_name))) {
     mtevL(noit_error,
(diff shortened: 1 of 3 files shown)


src/noit_filters.c
```diff
@@ -255,6 +257,20 @@ noit_filter_exists(const char *name) {
...
 }
+
+int
+  noit_filter_get_seq(const char *name, int64_t *seq) {
+  int exists;
+  void *v;
+  LOCKFS();
+  exists = mtev_hash_retrieve(filtersets, name, strlen(name), &v);
+  UNLOCKFS();
+  if(exists) {
```
postwait: Change to if(exists && seq != NULL) to better match convention in the rest of mtev/noit.

If the pass NULL in it will not crash and just return exists correctly.

src/noit_filters_rest.c
```diff
@@ -222,13 +246,18 @@ rest_set_filter(mtev_http_rest_closure_t *restc,
...
+  exists = noit_filter_get_seq(pats[1], &old_seq);
+  if(!node && exists == mtev_true) {
     /* It's someone else's */
     error_code = 403;
     goto error;
   }
 
-  if((newfilter = validate_filter_post(indoc, pats[1])) == NULL) goto error;
+  if((newfilter = validate_filter_post(indoc, pats[1], &seq)) == NULL) goto error;
+  if(exists && old_seq > seq) {
```
postwait: I think this should be

if(exists && ((old_seq >= seq) || (seq == 0)) {

So that you can't replace a filter without bumping the sequence number up -- unless seq is just zero and stays zero.  I thought that was how checks worked.

JonasKunze: I can't find this logic implemented for checks. That's what I found:
https://github.com/JonasKunze/reconnoiter/blob/master/src/noit_check_rest.c#L959

Should I still change it to your suggestion?

postwait: I think we should change both.  If the system uses sequences they must increase... This would allow one to replace something with the same sequence and then we wouldn't know to replicated it leaving us in an inconsistent state.

JonasKunze: I've changed the behavior for check here: https://github.com/circonus-labs/reconnoiter/pull/325

I didn't add the (seq == 0) because I don't see what it's supposed to do, as old_seq is always >=0, isn't it? I've enforced this now by changing it's type to uint64_t.

PR #336, opened 2016-11-08
Leverage CN better when adding/deleting noits.

postwait:
If the users specifies as CN when deleting, we should sweep
all noits (one) with the CN regardless of address.  When we add
a noit, if CN is specified, we should ensure no other registered
noit has that CN.  When we initiate connections we should use the
CN if available so as not to start duplicate sessions to common
target addresses (particular to mtev_reverse connections that have
a blank target).


rileyberton:
Looks good to me

PR #337, opened 2016-11-08
metric_director dedupes incoming messages

JonasKunze:
The timeout is 10s. Windows are cleared every 5s -> a duplicate is found if two messages are arriving within 10s up to 15s depending on the timing of the prune event

### DIFF
--- a/src/noit_metric_director.c
+++ b/src/noit_metric_director.c
@@ -46,6 +46,11 @@
 struct fq_conn_s;
 struct fq_msg;
 
+struct hash_and_time {
+  mtev_hash_table hash;
+  uint32_t last_touched_s;
+};
+
 static __thread struct {
   int id;
   ck_fifo_spsc_t *fifo;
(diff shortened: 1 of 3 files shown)


src/noit_metric_director.c
```diff
@@ -189,6 +198,86 @@ distribute_check(noit_metric_message_t *message) {
...
+    *stored_ts = whence;
+    if (mtev_hash_store(&dedupe_hashes, (const char *)stored_ts, sizeof(*stored_ts), hash_with_time) == 0) {
+      /* ugh, someone beat us */
+      free(stored_ts);
+      mtev_hash_destroy(hash, NULL, NULL);
+      free(hash_with_time);
+      if (mtev_hash_retrieve(&dedupe_hashes, (const char *)&whence, sizeof(whence), (void **)&hash_with_time) == 0) {
+        return NULL;
+      }
+    }
```
rileyberton: needs: `hash = &hash_with_time->hash` after the retrieve on 221

JonasKunze: right, fixed that!

PR #340, opened 2016-11-08
NULL is not a type, it is a value. Eradicate this.

postwait:


HeinrichHartmann:
This all looks plausible to me.

But I am not sure what problem this PR is addressing? 
Also, it is clear that no NULL_TYPES will be passed from "upstream"?

postwait:
There is no "null" type in the upstream.  The M record format prevents it.

PR #341, opened 2016-11-11
introduce concept of null values

postwait:


src/noit_metric_rollup.c
```diff
@@ -38,56 +38,67 @@
...
+      case METRIC_UINT32: return (double)v->value.v_uint32;
+      case METRIC_INT64: return (double)v->value.v_int64;
+      case METRIC_UINT64: return (double)v->value.v_uint64;
+      case METRIC_DOUBLE: return v->value.v_double;
+      default: ;
+    }
   }
-  return private_nan;
+  mtevFatal(mtev_error, "failed metric_value_double(%x, null: %d)\n",
+            v->type, v->is_null);
```
postwait: I think this change is too severe.  Callers could be relying on nan behavior.

pamaddox: Yeah.... may want to just return private_nan for now and revisit.

pamaddox: Maybe log it so we know when it happens?

PR #342, opened 2016-11-21
Monitoring message stream in metric_director

JonasKunze:
Added two atomics couning incoming and outgoing messages in metric_director


rileyberton:
Looks good to me

PR #347, opened 2016-12-16
B batches

postwait:


postwait:
clapper confirmed that the java readers can cope with this.

rileyberton:
Looks good to me

PR #351, opened 2017-01-24
CIRC-1134: Default to 256K for external check output reading

rileyberton:
Also provide a config variable "max_output_len" which will let this be increased to an arbitrary size
Also add a new metric to the check output which indicates if the output was truncated ala http checks.


postwait:
Lgtm

PR #354, opened 2017-02-02
support a standalone lua modules for libnoit

postwait:


postwait:
I also think this covers the intent of #350 so that one can be closed if this is accepted (or refactored and minimized)

src/modules/noit_lua_noit_binding.c
```diff
@@ -106,214 +105,6 @@ lua_general_filtersets_cull(lua_State *L) {
 }
 
 static int
-lua_noit_metric_adjustsubscribe(lua_State *L, short bump) {
```
gallison93: The library at the end of the file still references these deleted functions, which causes a compilation error. Lines 142-148.

postwait: missing file... i include the other C file that was missing that has these definitions.

src/modules/noit_lua_libnoit_binding.c
```diff
@@ -0,0 +1,263 @@
...
+  return 1;
+}
+
+static int
+lua_noit_metric_messages_distributed(lua_State *L) {
+  lua_pushinteger(L, noit_metric_director_get_messages_distributed());
+  return 1;
+}
+
+#ifndef NO_LUAOPEN_LIBNOIT
```
gallison93: I'd add a comment here directing the reader to the #define and #include at the top of `noit_lua_noit_binding.c`, to keep them aware of the dependency if they change this code. It is a simple dependency and the files are next to each other, but it would be an annoying issue to miss and have to debug.

PR #357, opened 2017-02-21
decoder: search forward for \n in histogram value_str

gallison93:
`payload`(and hence `value_str`) can have multiple newlines in the garbage between `payload[payload_len]` and the null-terminator. In such a case, scanning backwards for the first non-newline can leave garbage on the end of `metric->value.v_string` after the call to `mtev__strndup()`.

This change finds `vstrlen` by scanning forward for the first newline instead.

### DIFF
--- a/src/noit_message_decoder.c
+++ b/src/noit_message_decoder.c
@@ -73,7 +73,7 @@ int noit_message_decoder_parse_line(const char *payload, int payload_len,
     const char **noit_name, int *noit_name_len,
     noit_metric_value_t *metric, int has_noit) {
   const char *cp, *metric_type_str, *time_str, *check_id_str;
-  char *value_str;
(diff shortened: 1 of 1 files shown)


postwait:
This seems like a bug in whomever set payload_len. Payload length should be precise. Where is the payload coming from?

gallison93:
In the use case where this bug arose, `payload_len` is coming from the libmtev fq hook parameter of the same name. Last I checked, it was accurate, but I'll double check.

gallison93:
Yes, the passed `payload_len` is correct(== length of `payload` up to and including the first newline).

postwait:
The new line isnt required. And it often isn't null terminated. So let's
make sure and "cutting that short" is done with both those things in mind.

PR #358, opened 2017-03-20
Consul check

rileyberton:
User supplies the Consul health check url (e.g.: http://host:8500/v1/health/state/critical) and optionally the ServiceName and CheckName she cares about.  We generate mostly text metrics from the consul output, flattening into metric names as we go.

The blacklist params only apply to 'state' based health checks
The user can also supply a datacenter name to have that passed to the consul REST call.

### DIFF
--- a/src/modules-lua/noit/module/resmon/consul.lua
+++ b/src/modules-lua/noit/module/resmon/consul.lua
@@ -181,6 +181,20 @@ function first_to_upper(str)
     return (str:gsub("^%l", string.upper))
 end
 
+
+function sort_checks_by_name(tbl, sort_func)
+  local keys = {}
+  for key in pairs(tbl) do
+    table.insert(keys, key)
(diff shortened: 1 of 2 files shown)


src/modules-lua/noit/module/resmon/consul.lua
```diff
@@ -0,0 +1,388 @@
...
+function first_to_upper(str)
+    return (str:gsub("^%l", string.upper))
+end
+
+function json_to_metrics(check, doc, blacklists)
+    local services = 0
+    check.available()
+    local data = doc:document()
+    if data ~= nil then
+       if string.find(check.config.url, "/health/node") then
```
rileyberton: I was poorly attempting future compatibility (Assuming they don't change the name of the endpoints in `v2`.. This seems like a stupid idea.. will fix.

PR #360, opened 2017-03-27
Set "null" cloudwatch values to 0

neophenix:
Cloudwatch is super annoying in that if they don't have a value for
something, they simply don't return it.

I haven't looked at every single metric, but I think in most every
case this would mean the value is 0 (like requests, etc) because if
there was a value, it would be returned.

So, when we get back a datapoints section that is empty, loop over
the stats we requested, and return a 0 value instead of nothing for
that metric

### DIFF
--- a/src/modules-lua/noit/AWSClient.lua
+++ b/src/modules-lua/noit/AWSClient.lua
@@ -386,8 +386,21 @@ function AWSClient:perform(target, cache_table)
               --billing
               if count == 1 or namespace == "AWS/Billing" then
                 cache_table[metric] = most_recent
-              else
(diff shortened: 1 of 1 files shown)


neophenix:
Thought about the config option, wouldn't be hard.  

However, the primary reason for this change was to make these act in the expected manner, where we always get a value of some sort every run (on a normal OK run).  So allowing someone to break that expectation seemed bad?

That being said if we find their are issues with doing it this way, then we will have to probably add that option.

postwait:
I support making this the new default behavior without a config option; less complicated code.

PR #364, opened 2017-05-08
Make Truncated A Standalone Metric For External

pamaddox:
The external check should report "truncated" as a stand-alone metric
rather than making it part of the output string; putting it in the
output string can break parsing regexes.

This will cause issues if there is a configured metric called
"truncated", but that's better than the previous solution.


rileyberton:
Looks good to me

PR #366, opened 2017-05-20
Do not include \n in the string value of M records.

postwait:
The M format is \n terminated and the \n is not supposed to be part
of the value itself.  The vlen of th string included the \n and it
was being copied during the M record decoding process.


pamaddox:
Tested latest change, looks good.

PR #369, opened 2017-05-25
Safe watch real checks

postwait:


src/noit_check.c
```diff
@@ -981,7 +981,14 @@ noit_check_watch(uuid_t in, int period) {
 
   mtevL(noit_debug, "noit_check_watch(%s,%d)\n", uuid_str, period);
   if(period == 0) {
-    return noit_poller_lookup(in);
+    uuid_copy(n.checkid, in);
```
pamaddox: Use `mtev_uuid_copy`, it's much faster.

postwait: There are 17 uuid_copy's in the code, perhaps we can change them in an unrelated commit. This is just moving existing code around.

pamaddox: That's reasonable. Changed to an approval.

PR #373, opened 2017-06-21
This is changes to copyright claims to new dates.

SamuelWM01:


postwait:
The copyrights that were a single date should be appended to instead of changed. 

Copyright 2013

Should become

Copyright 2013-2017

SamuelWM01:
If the copyrights already have an extension such as
Copyright 2007-2013
Should I just change the last two digits or do something else?

postwait:
Yes. 2010-2013 should become 2013 to 2017

SamuelWM01:
I fixed my mistake but misnamed the commit as I must have accidentally wrote something in vim. Should be fine otherwise.

SamuelWM01:
also noted I fixed copyright extensions as (for example)
(initial) Copyright 2007-2013
as 
(edit) Copyright 2007-2017
I can go back and change that to 
Copyright 2013-2017
if need be

SamuelWM01:
Note
Found a file with an incorrect copyright (noit_check.c) its copyright has an extra character and says 20152-17. Need to determine correct range and update, was marked for later.

SamuelWM01:
should I add whole copyright prompt files missing a copyright completely? example of the prompt is the whole first comment in (noit_check.c).

PR #380, opened 2017-08-01
Jmx object filters

neophenix:
allow users to set mbean_properties_(.+) that will act like an OR filter for mbean objects, if they match the props, let them through.

This means we don't have to pull every object and pass it back to noit, potentially killing it while it tries to parse the large XML docs that can be created.

### DIFF
--- a/src/java/src/com/omniti/jezebel/check/jmx.java
+++ b/src/java/src/com/omniti/jezebel/check/jmx.java
@@ -33,6 +33,7 @@
 package com.omniti.jezebel.check;
 
 import java.util.HashMap;
+import java.util.Hashtable;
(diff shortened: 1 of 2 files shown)


postwait:
Do we need an update to the jmx xml to state that these propoerties are allowed? (docs update)

neophenix:
Yeah, where are the docs for these located and I'll do that

postwait:
Shit. Doesn't look like we have docs for this.   But adding a jmx.xml in the modules/ directory would allow for docs.  If you make it look like any of the other xml files, we can autogen docs off it.

neophenix:
OK, I think that is what you meant

PR #381, opened 2017-08-01
more reconnoiter coverity fixes

rileyberton:


src/noit_check_log_helpers.c
```diff
@@ -468,8 +468,6 @@ noit_check_log_bf_to_sm(const char *line, int len, char ***out, int noit_ip)
 
  bad_line:
   if(*out) {
-    int i;
-    for(i=0; i<cnt + has_status; i++) if((*out)[i]) free((*out)[i]);
```
postwait: I think this leaks now.  (*out)[i] is malloc'd above.

rileyberton: This appears to be totally broken, we always appear to free `*out` which is counter to the point of this function.  Fixing.

PR #382, opened 2017-08-07
Fix Memory Leak When Compression Disabled

pamaddox:
We had an instance of inverted logic; we need to free this buffer if
compression is disabled, not the other way around.


aidancully:
LGTM

PR #383, opened 2017-08-07
Fix Memory Leak In External Check

pamaddox:
We were never cleaning up the stdoutbuff or stderrbuff in the external
check, leading to a memory leak. Fix it.


src/modules/external.c
```diff
@@ -475,6 +475,8 @@ static int external_handler(eventer_t e, int mask,
     ci->error = data->cr->stderrbuff;
     ci->stdout_truncated = data->cr->stdout_truncated;
     ci->stderr_truncated = data->cr->stderr_truncated;
+    free(data->cr->stdoutbuff);
+    free(data->cr->stderrbuff);
```
aidancully: `ci->output` and `ci->error` point to this memory, so it seems wrong to free it...

pamaddox: OK

PR #386, opened 2017-08-09
Google Analytics v4 api plugin

maier:


src/modules-lua/noit/module/googleanalytics4.lua
```diff
@@ -0,0 +1,565 @@
...
+        oauth_token_secret = check.config.oauth_token_secret,
+        table_id           = check.config.table_id,
+        metric_group       = check.config.metric_group,
+        oauth_version      = check.config.oauth_version,
+        metrics            = {}
+    }
+
+    local uuid         = check.checkid
+    local table_hit    = cache_table[uuid]
+    local current_time = os.time()
```
postwait: use `mtev.gettimeofday`

src/modules-lua/noit/module/googleanalytics4.lua
```diff
@@ -0,0 +1,565 @@
...
+                check.metric_uint64(name, value)
+            end
+        else
+            check.metric_string(name, value)
+        end
+        metric_count = metric_count + 1
+    end
+
+    -- Check about every 15 minutes
+    if current_time - cache_table[uuid]['timestamp'] >= 890 then
```
postwait: 890 isn't 15 minutes, also shouldn't this be `current_time - 60*15`

maier: not sure, this is from the original plugin. will update to reflect 15 minutes.

src/modules-lua/noit/module/googleanalytics4.lua
maier: ok

src/modules-lua/noit/module/googleanalytics4.lua
```diff
@@ -0,0 +1,565 @@
...
+                check.metric_uint64(name, value)
+            end
+        else
+            check.metric_string(name, value)
+        end
+        metric_count = metric_count + 1
+    end
+
+    -- Check about every 15 minutes
+    if (current_time - cache_table[uuid]['timestamp']) >= (current_time - (60 * 15)) then
```
postwait: Isn't the `cache_table[uuid]['timestamp']` the number of seconds since epoch?

If so, you'd want current_time only on one side.

I would think something like:

` if ( cache_table[uuid]['timestamp'] + (60 * 15) >= current_time ) then `

PR #392, opened 2017-08-28
Fix A Bunch Of Memory Leaks

pamaddox:


src/modules/check_test.c
```diff
@@ -204,8 +204,12 @@ rest_test_check_result(struct check_test_closure *cl) {
 
   mtevL(nlerr, "Flushing check test result\n");
 
-  if(cl->restc->call_closure_free)
+  if(cl->restc->call_closure_free) {
     cl->restc->call_closure_free(cl->restc->call_closure);
+    /* We have to free this manually because the closure set by
+     * rest_get_xml_upload doesn't do it */
+    free(cl->restc->call_closure);
```
aidancully: seems we should fix `rest_get_xml_upload`, instead?

pamaddox: I can look into that, but it's riskier. Lots of other stuff depends on that function and it's an external library call.

aidancully: It's my strong opinion that we should fix the free function set by `rest_get_xml_upload`... The abstraction boundary is such that anything that sets the `call_closure` should be the thing that knows how to free the `call_closure`.

pamaddox: OK, fixed in libmtev and reverted calls to manually fix it here.

PR #393, opened 2017-08-29
Fix Lock Leak In noit_module_init

pamaddox:
We were failing to release all of the conf locks we were acquiring in
noit_module_init.... fix to release them properly.


aidancully:
@gallison93 identified another error, we should fix that as well, but this one LGTM

PR #394, opened 2017-08-29
Add A Fix To A Potential Console Deadlock

pamaddox:


src/noit_filters.c
```diff
@@ -546,6 +546,8 @@ noit_console_rule_configure(mtev_console_closure_t ncct,
                       sec2 = MTEV_CONF_EMPTY,
                       byebye = MTEV_CONF_EMPTY;
   mtev_conf_t_userdata_t *info;
+  mtev_conf_section_t add_arg;
```
rileyberton: Can just do: `mtev_conf_section_t agg_arg = NULL;` and lose the next line

src/noit_filters.c
```diff
@@ -588,6 +589,7 @@ noit_console_rule_configure(mtev_console_closure_t ncct,
       int rulenum = atoi(argv[1]);
       snprintf(xpath, sizeof(xpath), "rule[%d]", rulenum);
       add_arg = mtev_conf_get_section(fsnode, xpath);
+      add_arg_initialized = mtev_true;
```
rileyberton: Can lose this

src/noit_filters.c
```diff
@@ -623,6 +625,9 @@ noit_console_rule_configure(mtev_console_closure_t ncct,
   }
   rv = 0;
  bail:
+  if (add_arg_initialized) {
```
rileyberton: Can lose the wrapping `if`

pamaddox: Already merged this, but I'll go ahead and make these changes/push to master.

gallison93: `add_arg` might have been set to `fsnode`, if the release(of both `add_arg` and `fsnode`) is unconditional it seems like this function will unlock an extra time

pamaddox: Yeah, that's why I did it this way.

I think it's probably safest to just leave it like this for now... if we rearrange, I'll do a separate PR.

PR #395, opened 2017-08-29
Fix Memory Leak In Httptrap Module

pamaddox:
1) Don't allocate the call closure until we're past all of the validity
   checks
2) If we get into the error condition, we need to clean up properly.


pamaddox:
OK, removed that bit.

PR #396, opened 2017-08-29
Fix DNS Memory Leak

pamaddox:
We need to eventer_remove_fde and then eventer_free the events in the
dns check. Otherwise, we leak memory.


pamaddox:
This doesn't work in some circumstances. Need to rethink it. Withdrawing.

PR #397, opened 2017-08-29
Fix DNS Memory Leak

pamaddox:
We were leaking memory on bad DNS checks; need to free the eventer stuff
manually if we're not coming from an eventer callback.

### DIFF
--- a/src/modules/dns.c
+++ b/src/modules/dns.c
@@ -92,6 +92,17 @@ static void dns_module_dns_ctx_handle_free(void *vh) {
   mtevAssert(h->timeout == NULL);
   free(h);
 }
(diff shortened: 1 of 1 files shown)


src/modules/dns.c
```diff
@@ -92,6 +92,17 @@ static void dns_module_dns_ctx_handle_free(void *vh) {
   mtevAssert(h->timeout == NULL);
   free(h);
 }
+static void dns_module_dns_ctx_handle_free_and_remove_eventer(void *vh) {
+  dns_ctx_handle_t *h = vh;
```
rileyberton: Should check for NULL on `vh` and return if NULL

pamaddox: Fixed

PR #398, opened 2017-09-01
Fix Connect Error In SSH Check

pamaddox:
The logic for setting a timeout in the SSH check was totally busted - we were setting a default timeout of 10ms every time since we could never get into the code that sets it.

Change it for now to remove broken code altogether and give it a larger timeout.


pamaddox:
I'm going to do this a bit differently (IE, not just a high hardcoded value). Closing.

PR #399, opened 2017-09-01
Fix Bug In SSH2 Check Connection Timeout

pamaddox:
We were always setting the connection timeout to 10ms, which is wrong.
The logic for checking the timeout was busted. Fix to store the actual
timeout data in the check's info struct and use that.


aidancully:
LGTM

PR #401, opened 2017-09-06
Fix Memory Leaks Related To Websockets And Decompression

pamaddox:
There were a couple of significant memory leaks - one related to not
freeing metric data on streaming and one related to not cleaning up
compression data. Fixed both of these.


postwait:
Lgtm

PR #405, opened 2017-10-17
Modernize lua ssh2 module

esproul:
Enables the ssh2 module to check a wider variety of more modern SSH servers.  Updates include:
 * SHA256 fingerprint metric, and base64-encoded versions of both SHA1 and SHA256 fingerprint metrics, matching what newer ssh clients display.
 * New host key types: ECDSA, Ed25519
 * Additional, larger DH groups from RFC 3526 for key exchange
 * Modern ciphers, including ChaCha20-Poly1305 and AES-GCM
 * Newer MAC algorithms


esproul:
Should also mention, the SHA256 fingerprint support requires a libmtev that includes https://github.com/circonus-labs/libmtev/pull/327

src/modules-lua/noit/module/ssh2.lua
```diff
@@ -48,23 +48,23 @@ function onload(image)
     <parameter name="method_kex"
                required="optional"
                default="diffie-hellman-group14-sha1"
-               allowed="^diffie-hellman-(?:group1-sha1|group14-sha1|group-exchange-sha1)$">The key exchange method to use.</parameter>
```
postwait: You've dropped group-exchange-sha1 ?

src/modules-lua/noit/module/ssh2.lua
```diff
@@ -48,23 +48,23 @@ function onload(image)
...
+               allowed="^diffie-hellman-(?:group1-sha1|group14-sha1|group16-sha512|group18-sha512)$">The key exchange method to use.</parameter>
     <parameter name="method_hostkey"
                required="optional"
                default="ssh-rsa"
-               allowed="^(?:ssh-dss|ssh-rsa)$">The host key algorithm supported.</parameter>
+               allowed="^(?:ssh-rsa|ecdsa-sha2-nistp256|ssh-ed25519|ssh-dss)$">The host key algorithm supported.</parameter>
     <parameter name="method_crypt_cs"
                required="optional"
-               allowed="^(?:aes256-ctr|aes192-ctr|aes128-ctr|aes256-cbc|aes192-cbc|aes128-cbc|blowfish-cbc|arcfour128|arcfour|cast128-cbc|3des-cbc|none)$">The encryption algorithm used from client to server.</parameter>
```
postwait: why are these removed?

src/modules-lua/noit/module/ssh2.lua
```diff
@@ -117,30 +117,37 @@ function elapsed(check, name, starttime, endtime)
...
 local ENCALG = {
-  'aes128-cbc', 'aes192-cbc', 'aes256-cbc', 'aes128-ctr', 'aes192-ctr',
-  'aes256-ctr', 'rijndael128-cbc', 'rijndael192-cbc', 'rijndael256-cbc',
-  '3des-cbc', '3des-ecb', '3des-cfb', '3des-ofb', '3des-ctr',
-  'blowfish-cbc', 'blowfish-ecb', 'blowfish-cfb', 'blowfish-ofb',
-  'blowfish-ctr', 'twofish128-ctr', 'twofish128-cbc', 'twofish192-ctr',
-  'twofish192-cbc', 'twofish256-ctr', 'twofish256-cbc', 'twofish-cbc',
-  'twofish-ecb', 'twofish-cfb', 'twofish-ofb', 'cast128-cbc', 'cast128-ecb',
-  'cast128-cfb', 'cast128-ofb', 'idea-cbc', 'idea-ecb', 'idea-cfb',
-  'idea-ofb', 'arcfour', 'arcfour128', 'arcfour256' }
```
postwait: again why are things removed instead of just added to here?

src/modules-lua/noit/module/ssh2.lua
esproul: Yes, because it doesn't work.  If you think it should work, I can have another look, but I'm fairly certain I tested it and it failed.

src/modules-lua/noit/module/ssh2.lua
esproul: I was cleaning up the list of rarely-used options.  Note that I did preserve all the aes*-{ctr,cbc}.

src/modules-lua/noit/module/ssh2.lua
esproul: I removed all of the ones that use obscure modes (modes other than CTR or CBC).  I have never seen an SSH server in the wild that offers them, and ciphers like CAST and IDEA are flat-out obsolete in 2017.

src/modules-lua/noit/module/ssh2.lua
esproul: From my understanding of RFC 4419, if we want to do `diffie-hellman-group-exchange-*` we have to construct a different kexinit packet that lists our min/max/preferred group size and let the server choose the group.  I wasn't prepared to go down that particular rabbit hole.

esproul:
Per out-of-band discussion, I'll restore all old KEX methods, ciphers, and MAC algos. Since we are a tester, we should allow even old, obsolete options to ensure the widest array of targets.

esproul:
Restoring removed bits does raise the question of which is authoritative, the module's XML description of allowed values, or the lists of methods, ciphers, etc. in the actual code?  For instance, in the existing code, not all of the entries in `ENCALG` are allowed in the XML.

esproul:
@postwait I believe I've addressed your concerns about removing options.

PR #407, opened 2017-10-23
Never iterate starting at -1.

postwait:
If the XML and hash didn't agree, we could never set n = 0 and then
start filling in the array at -1.  This is simplified and we now
treat surgery on the XML separate from in-memory pruning.


aidancully:
LGTM

PR #408, opened 2017-10-25
Extend the metric decoding to extract stream and measurement tags.

postwait:
* Adopt tagset stuff from aidan with mild API variance.
 * enforce tag size limits

### DIFF
--- a/src/noit_message_decoder.c
+++ b/src/noit_message_decoder.c
@@ -48,6 +48,51 @@
     } \
 } while(0)
 
+mtev_boolean
+noit_metric_extract_tags(const char *in, int *inlen,
+                         const char *start_marker, char end_marker,
+                         noit_metric_tagset_builder_t *builder) {
+  /* This function will take a metric name (in) and extract a
+   * <start_marker><taglist><end_marker> from it
+   * and union that with any existing tags.
+   * If successful, it will shorten inlen accordingly.
+   */
+  int st_off = strlen(start_marker);
+  const char *end = in + *inlen - 1;
+  const char *match = strnstrn(start_marker, st_off, in, *inlen);
+  /* we wannt to find the last one */
+  const char *tag_start = match;
+  while(match) {
+    match = strnstrn(start_marker, st_off,
+                     tag_start+1, (*inlen) - (tag_start + 1 - in));
+    if(match) tag_start = match;
+  }
+  if(!tag_start || *end != end_marker) return mtev_false;
+  if(tag_start + st_off == end) { /* empty */
+    *inlen -= st_off + 1;
+    return mtev_true;
+  }
+
+  if(!noit_metric_tagset_builder_add(builder, tag_start + st_off, end - (tag_start + st_off)))
+    return mtev_false;
+
+  *inlen = tag_start - in;
+  return mtev_true;
+}
+void
+noit_metric_process_tags(noit_metric_message_t *metric) {
+  noit_metric_tagset_builder_t stream_builder, measurement_builder;
+  noit_metric_tagset_builder_start(&stream_builder);
+  noit_metric_tagset_builder_start(&measurement_builder);
+  while(
+    noit_metric_extract_tags(metric->id.name, &metric->id.name_len,
+                             "|ST[", ']', &stream_builder) ||
+    noit_metric_extract_tags(metric->id.name, &metric->id.name_len,
+                             "|MT{", '}', &measurement_builder)
+  );
+  noit_metric_tagset_builder_end(&stream_builder, &metric->id.stream, NULL);
+  noit_metric_tagset_builder_end(&measurement_builder, &metric->id.measurement, NULL);
+}
 int noit_is_timestamp(const char *line, int len) {
   int is_ts = 0;
   int state = 0, cnt[2] = { 0, 0 };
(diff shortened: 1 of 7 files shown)


src/noit_message_decoder.c
```diff
@@ -174,32 +219,356 @@ int noit_message_decoder_parse_line(const char *payload, int payload_len,
...
+   * anyway... helps our display logic. */
+  char *canonical = malloc(canonical_size + 1);
+  if(!canonical) return NULL;
+  if(noit_metric_tags_canonical(tags, tag_count, canonical, canonical_size) != canonical_size) {
+    free((void *) canonical);
+    return NULL;
+  }
+  canonical[canonical_size] = '\0';
+  /* reparse the tags, so that they point into the canonical string.
+   * (this simplifies life-time management for the lookup_t
```
aidancully: this comment is out-of-date in the original... should be `for the noit_metric_tagset_t structure`.

src/noit_message_decoder.c
```diff
@@ -174,32 +219,356 @@ int noit_message_decoder_parse_line(const char *payload, int payload_len,
...
+   * structure.) */
+  noit_metric_tags_parse(canonical, canonical_size, tags, tag_count, &canonical_size);
+
+  return canonical;
+}
+
+int
+noit_metric_tagset_init(noit_metric_tagset_t *lookup,
+                        const char *tagnm, size_t tagnmlen) {
```
aidancully: `tagnm` (and `tagnmlen`) is the wrong var-name here, implies that it's the name of a single tag, where this function expects 0-or-more. `tagsetstr`/`tagsetstrlen`?

src/noit_metric.h
```diff
@@ -77,9 +77,23 @@ typedef enum {
 } noit_message_type;
 
 typedef struct {
+  uint16_t total_size;
+  uint8_t category_size;
+  const char *tag;
+} noit_metric_tag_t;
+
+typedef struct noit_metric_tagset_t {
+  noit_metric_tag_t *tags;
```
aidancully: life-cycle management of this structure is complicated, since `tags` point into user-supplied memory, so the source memory must outlive this structure...

src/noit_message_decoder.c
```diff
@@ -174,32 +219,356 @@ int noit_message_decoder_parse_line(const char *payload, int payload_len,
...
+    return 0;
+  }
+  noit_metric_tag_t *tags = (noit_metric_tag_t *) calloc(tag_count, sizeof(noit_metric_tag_t));
+  size_t canonical_size;
+  ssize_t parsed_tag_count = noit_metric_tags_parse(tagnm, tagnmlen, tags, tag_count, &canonical_size);
+  if(parsed_tag_count <= 0) {
+    free((void *) tags);
+    return -1;
+  }
+  lookup->tags = tags;
```
aidancully: suggest `lookup->tags = realloc(tags), tag_count * sizeof(noit_metric_tag_t *)` here. number of elements in `tags` array might have shrunk since initial allocation, and I believe `realloc` is fast when reducing size.

src/noit_message_decoder.c
```diff
@@ -48,6 +48,51 @@
...
+                     tag_start+1, (*inlen) - (tag_start + 1 - in));
+    if(match) tag_start = match;
+  }
+  if(!tag_start || *end != end_marker) return mtev_false;
+  if(tag_start + st_off == end) { /* empty */
+    *inlen -= st_off + 1;
+    return mtev_true;
+  }
+
+  if(!noit_metric_tagset_builder_add(builder, tag_start + st_off, end - (tag_start + st_off)))
```
aidancully: `builder_add` adds a single tag at a time, while the data inside `|ST[<tagset>]` is multiple tags, so I don't think this works.

src/noit_message_decoder.c
```diff
@@ -174,32 +219,356 @@ int noit_message_decoder_parse_line(const char *payload, int payload_len,
...
+0,0,0,0,0,0,0,0,0,0,0,0,0,1,1,0,1,1,1,1,1,1,1,1,1,1,0,0,0,0,0,0,
+0,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,0,0,0,1,
+0,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,0,0,0,0,
+0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
+0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
+0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
+0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
+};
+mtev_boolean
+noit_metric_tagset_is_taggable_char(char c) {
```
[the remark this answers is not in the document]
postwait: it is on the next line.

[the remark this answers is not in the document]
aidancully: Oops, sorry :-)

src/noit_message_decoder.c
```diff
@@ -48,6 +48,51 @@
...
+  if(!tag_start || *end != end_marker) return mtev_false;
+  if(tag_start + st_off == end) { /* empty */
+    *inlen -= st_off + 1;
+    return mtev_true;
+  }
+
+  if(!noit_metric_tagset_builder_add(builder, tag_start + st_off, end - (tag_start + st_off)))
+    return mtev_false;
+
+  *inlen = tag_start - in;
```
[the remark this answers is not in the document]
postwait: It looks for the last and because they are differently bracketed it works.
|MT{....}|ST[...] won't match at all because MT{ doesn't match the ending ].
It then cycles pulling at least one off the end until it cannot successfully pull either.

src/noit_message_decoder.c
```diff
@@ -174,32 +219,356 @@ int noit_message_decoder_parse_line(const char *payload, int payload_len,
...
+    else {
+      /* expect a valid tag character */
+      if(!noit_metric_tagset_is_taggable_char(test_char)) return 0;
+    }
+    cur_size++;
+  }
+  /* make sure we covered everything */
+  if(colon_pos == 0 || colon_pos >= cur_size - 1) return 0;
+  /* tag category and name must both be <= 127 */
+  if(colon_pos > 127 || cur_size - colon_pos - 1 > 127) return 0;
```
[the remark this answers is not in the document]
postwait: agreed

src/noit_metric.h
postwait: That's why I dropped the forced canonicalization.. The caller here already has memory it is ensuring that will last longer than the tagset, so we'd like to avoid the addition allocation.

aidancully:
I think the snowth metric-name parsing is more robust than what we have here? (it'd accept but ignore `|UNKNOWN{}` metadata, and reject `|UNKNOWN[]` metadata), but this looks like it will work.

PR #421, opened 2018-01-18
API for constructing 'complex' tag matching criteria

postwait:
* API for string -> AST
 * sample implementation of matching an AST against a single tagset.

### DIFF
--- a/src/noit_metric_tag_search.h
+++ b/src/noit_metric_tag_search.h
@@ -1,0 +1,102 @@
+/*
+ * Copyright (c) 2018, Circonus, Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *     * Redistributions of source code must retain the above copyright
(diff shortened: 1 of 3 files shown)


src/noit_metric_tag_search.h
```diff
@@ -0,0 +1,102 @@
...
+ * stmt-list   :=  stmt-single,stmt-list
+ *              |  stmt-single
+ * stmt-single :=  stmt
+ *              |  match
+ * match       :=  partmatch
+ *              |  partmatch : partmatch
+ * partmatch   :=  [a-zA-Z0-9._-*]  // tag character plus wildcard
+ *              |  /<re>/           // re is a regular expression
+ *
+ * example: and(env:prod,or(service:web,team:/^(red|blue)$/),not(special))
```
rileyberton: I think the `not` clause will need to be a list

postwait: How does not apply to a list?  Traditionally not or ! in most languages is a unary operator.

rileyberton: yes, I see.. misread the grammar

PR #426, opened 2018-03-20
Put metric name into name attribute, not in the tag

HeinrichHartmann:
Fixing XML parsing errors when the metric name contains brackets e.g. `output[1]`


HeinrichHartmann:
right. This is fixed now. Here is how the resulting XML looks:
```
  <check uuid="3c94bf4a-c395-48e0-821e-de96faee5a0e" target="q._caql" module="caql" name="c_603_70005::caql" filterset="filterset_1445" period="60000" timeout="10000" seq="2965">
    <config>
      <query>histogram{1,2,4}</query>
      <histogram:value name="output[1]">add</histogram:value>
    </config>
  </check>
```

PR #430, opened 2018-03-27
Introduce replication_prefix attribute for filtersets

HeinrichHartmann:
This allows us to move replicated filtersets to a sub-node.


HeinrichHartmann:
Update the PR.

PR #431, opened 2018-04-04
Move "running on other node" info to debug/cluster log

HeinrichHartmann:


src/noit_check_tools.h
```diff
@@ -151,7 +151,7 @@ MTEV_HOOK_PROTO(check_postflight,
 
 #define BAIL_ON_RUNNING_CHECK(check) do { \
   if(!noit_should_run_check(check, NULL)) { \
```
postwait: static mtev_log_stream_t debug_cluster= NULL; \
if(!debug_cluster) debug_cluster = mtev_log_stream_find("debug/cluster"); \
mtevL(debug_cluster, "Check %s is running on another node\n", check->name); \

HeinrichHartmann:
Updated

src/noit_check_tools.h
```diff
@@ -149,9 +149,12 @@ MTEV_HOOK_PROTO(check_postflight,
                 void *, closure,
                 (void *closure, noit_module_t *self, noit_check_t *check, noit_check_t *cause))
 
+static mtev_log_stream_t noit_cluster_debug = NULL;
```
postwait: Put this in the if... it's not good practice to declare a static file variable in a header.  You want static function scope.

HeinrichHartmann:
yes, yes, yes. Just noted I pushed an old version. Fixed now. Sry.

HeinrichHartmann:
Sorry. I have one more question here, if I initialize the variable to NULL we will always do the lookup again. If I don't, I am not sure if it NULL the first time we see it.

postwait:
That's not how static initialization works.  That initialization only every happens once.

HeinrichHartmann:
Cool. Then I'll merge this as is. Thx for the explanation.

PR #433, opened 2018-04-23
Base64 tag decoding

rileyberton:


src/noit_metric.c
```diff
@@ -231,17 +234,63 @@ static uint8_t vtagmap[256] = {
...
 noit_metric_tagset_is_taggable_key(const char *key, size_t len)
 {
-  /* it's likely faster to just check all the chars than to have a branch and
-     decide to exit early */
+  /* there are 2 tag formats supported, plain old tags that obey the vtagmap
+     charset, and base64 encoded tags that obey the:
+
+     ^b"<base64 chars>"$ format
+  */
+  if (len >= 2) {
```
postwait: >= 3... might as well count the closing " as line 283 assumes it is there.

src/noit_metric.c
```diff
@@ -231,17 +234,63 @@ static uint8_t vtagmap[256] = {
...
+    /* must start with b" */
+    const char *x = strnstrn(key, 2, "b\"", 2);
+    if (x) {
+      /* and end with " */
+      const char *y = strnstrn(&key[len - 2], 1, "\"", 1);
+      if (y) {
+	size_t sum_good = 0;
+	for (size_t i = 2; i < len - 3 /* skip the wrapping chars */; i++) {
```
postwait: I think you still want < len-1 .... 2 in the front one in the back == 3

src/noit_metric.c
```diff
@@ -254,3 +303,48 @@ noit_metric_tagset_is_taggable_value(const char *val, size_t len)
...
+
+size_t
+noit_metric_tagset_decode_tag(char *decoded_tag, size_t max_len, const char *encoded_tag, size_t encoded_size)
+{
+  const char *colon = (const char *)strnstrn(encoded_tag, encoded_size, ":", 1);
+  if (!colon) return 0;
+
+  char *decoded = decoded_tag;
+  const char *encoded = encoded_tag;
+  const char *m = (const char *)strnstrn(encoded, 2, "b\"", 2);
```
postwait: if they are both two long, why is this not a memcmp?

src/noit_metric.c
```diff
@@ -254,3 +303,48 @@ noit_metric_tagset_is_taggable_value(const char *val, size_t len)
...
+  if (m) {
+    encoded += 2;
+    const char *eend = strchr(encoded, '"');
```
postwait: You want memchr here, encoded might not be terminated.

src/noit_metric.c
```diff
@@ -254,3 +303,48 @@ noit_metric_tagset_is_taggable_value(const char *val, size_t len)
...
+  }
+  else {
+    memcpy(decoded, encoded, colon - encoded);
+    decoded += (colon - encoded);
+    encoded += (colon - encoded);
+  }
+  encoded++; // skip over the colon
+  *decoded = 0x1f; //replace colon with ascii unit sep
+  decoded++;
+  m = (const char *)strnstrn(encoded, 2, "b\"", 2);
```
postwait: memcmp

src/noit_metric.c
```diff
@@ -254,3 +303,48 @@ noit_metric_tagset_is_taggable_value(const char *val, size_t len)
...
+  if (m) {
+    encoded += 2;
+    const char *eend = strchr(encoded, '"');
```
postwait: memchr

postwait: I think you also want to confirm that **end is indeed at the end.

PR #438, opened 2018-05-04
Make metric names canonical with respect to tags.

postwait:
* Expand tag characters to a large alowable set
 * Define form as name|ST[..]|MT{...}
 * |ST and |MT are omitted if no such tags are there
 * Base64 tags are decoded before sorting.
 * Tags within groups are unique and sorted.
 * Only tags parts (key or value) are base64 encoded if they cannot be expressed otherwise.
 * Canonicalize on the way into the set metrics path for checks.
 * Canonicalize on the message decoding path (extra alloc if non-canonical)

### DIFF
--- a/src/noit_check.c
+++ b/src/noit_check.c
@@ -2023,7 +2023,7 @@ cleanse_metric_name(char *m) {
     *cp = '\0';
 }
 
-int
+static int
 noit_stats_populate_metric(metric_t *m, const char *name, metric_type_t type,
                            const void *value) {
   void *replacement = NULL;
@@ -2036,6 +2036,12 @@ noit_stats_populate_metric(metric_t *m, const char *name, metric_type_t type,
 
   m->metric_name = strdup(name);
   cleanse_metric_name(m->metric_name);
+  if(noit_metric_canonicalize(m->metric_name, strlen(m->metric_name),
+                              m->metric_name, strlen(m->metric_name),
+                              mtev_true) <= 0) {
+    free(m->metric_name);
+    return -1;
+  }
 
   if(type == METRIC_GUESS)
     type = noit_metric_guess_type((char *)value, &replacement);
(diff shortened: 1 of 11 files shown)


src/noit_check.c
```diff
@@ -2109,9 +2122,16 @@ noit_stats_set_metric(noit_check_t *check,
 
 void
 noit_stats_set_metric_coerce_with_timestamp(noit_check_t *check,
-                             const char *name, metric_type_t t,
+                             const char *name_raw, metric_type_t t,
                              const char *v,
                              struct timeval *timestamp) {
+  char name[MAX_METRIC_TAGGED_NAME];
+  if(strlen(name_raw) > sizeof(name)-1) return;
```
rileyberton: This block is called more than once.  Make a function?

src/noit_message_decoder.c
```diff
@@ -334,17 +397,15 @@ noit_metric_tags_parse_one(const char *tagnm, size_t tagnmlen,
...
-  const noit_metric_tag_t *r = (noit_metric_tag_t *) v_r;
-  size_t memcmp_len = l->total_size < r->total_size ? l->total_size : r->total_size;
-  int cmp_rslt = memcmp(l->tag, r->tag, memcmp_len);
-  if(cmp_rslt != 0) return cmp_rslt;
-  if(l->total_size < r->total_size) return -1;
-  if(l->total_size > r->total_size) return 1;
-  return 0;
+tag_canonical_size(noit_metric_tag_t *tag) {
+  int len;
+  char dbuff[NOIT_TAG_MAX_PAIR_LEN], ebuff[NOIT_TAG_MAX_PAIR_LEN];
```
rileyberton: Don't need `ebuff` here since this is just throwaway decode/encode for sizing reasons.

src/noit_metric.c
```diff
@@ -345,3 +429,120 @@ noit_metric_tagset_decode_tag(char *decoded_tag, size_t max_len, const char *enc
...
+noit_metric_canonicalize(const char *input, size_t input_len, char *output, size_t output_len,
+                         mtev_boolean null_term) {
+  int i = 0, ntags = 0;
+  noit_metric_tag_t stags[MAX_TAGS], mtags[MAX_TAGS];
+  int n_stags = 0, n_mtags = 0;
+  char buff[MAX_METRIC_TAGGED_NAME];
+  if(output_len < input_len) return -1;
+  if(input != output) memcpy(output, input, input_len);
+
+  if(input_len > MAX_METRIC_TAGGED_NAME) return -1;
```
rileyberton: Reverse this line and the previous line.

src/noit_metric.h
```diff
@@ -77,6 +79,7 @@ typedef enum {
 } noit_message_type;
 
 #define NOIT_TAG_MAX_COMPONENT_LEN 127
+#define NOIT_TAG_MAX_PAIR_LEN 256
```
rileyberton: This might be too short.. Also, I would allow for shorter categories and longer values.

PR #447, opened 2018-05-30
Tools-only install target

### LINKED ISSUE #446: Want a tools-only install target
Would like to ship a separate package of add-on tools, of which there is currently only one: `noit_b2sm`.

esproul:
Right now this is just noit_b2sm but there may eventually be more.

Fixes #446


esproul:
I decided to leave it in existing targets (currently just `install-noitd-nolibs`) because we are not likely to co-install a future "noit-tools" package with the broker, and I figured it was best to leave the tool in place for brokers.

Happy to change that if desired.

postwait:
You're the packaging hero... if this is what you want that's fine.  Just wanted to make sure it wasn't an omission.

PR #448, opened 2018-06-01
Use metric name with tags to lookup interest

HeinrichHartmann:
Fixes stream tags fallout in caql-broker


src/noit_metric_director.c
```diff
@@ -240,7 +240,7 @@ distribute_metric(noit_metric_message_t *message) {
   if(mtev_hash_retrieve(&id_level, (const char *) &message->id.id, UUID_SIZE,
       &vhash)) {
     if(mtev_hash_retrieve((mtev_hash_table *) vhash, message->id.name,
-        message->id.name_len, &vinterests)) {
+        message->id.name_len_with_tags, &vinterests)) {
```
rileyberton: Is this the only place we need to use the longer len?

PR #449, opened 2018-06-04
Return "metric_name_with_tags" to lua

HeinrichHartmann:
This restores the behavior of returning the full metric name, instead
of truncating stram tag. In this way we keep the policy that a metric
is uniquely identified by UUID+name.


HeinrichHartmann:
The behavior of .name was suggested by @postwait:

> I'd add name_without_tags and make name drt wrt tags

PR #456, opened 2018-06-14
Support check:should_run_check()

postwait:


HeinrichHartmann:
Nit: I don't get why this is called "is_on_primary_cluster_node". What is the primary node? I thought all nodes are equal. The question is only: Should I run this check?
Why not keep the name "should_run_check" that we use in the C code?

HeinrichHartmann:
Otherwise LGTM.

PR #468, opened 2018-07-21
Address coverity nits

postwait:


postwait:
This needs to be tested... Specifically, the missing_namespaces function concerns me.  It wasn't doing what it was intended to do, but we need to make sure the original intentions are actually good.

PR #471, opened 2018-07-26
accept tags with a category but a blank value

rileyberton:


src/noit_metric.c
```diff
@@ -326,7 +326,10 @@ noit_metric_tagset_is_taggable_key(const char *val, size_t len)
 mtev_boolean
 noit_metric_tagset_is_taggable_value(const char *val, size_t len)
 {
-  return noit_metric_tagset_is_taggable_part(val, len, noit_metric_tagset_is_taggable_value_char);
+  /* accept blank string, blank base64 encoded string, and acceptable taggable_value_chars as values */
+  return len == 0 ||
+    memcmp("b\"\"", val, MIN(len,3)) == 0 ||
```
postwait: This will match `"b"` and `b"`...

You want `(len == 3 && memcmp("b\"\"", val, 3))`

rileyberton: ahh, yes, good catch

PR #484, opened 2018-08-22
support new mtev_lua_gc

postwait:


src/modules/lua_check.c
```diff
@@ -972,13 +974,12 @@ noit_lua_check_resume(mtev_lua_resume_info_t *ri, int nargs) {
         mtevL(nldeb, "lua_State(%p) -> %d [check: %p]\n", ri->coro_state,
               lua_status(ri->coro_state), ci ? ci->check: NULL);
       }
-      lua_gc(mtev_lua_lmc_L(ri->lmc), LUA_GCCOLLECT, 0);
```
cody271: Should have `mtev_lua_gc()` here ..?

postwait: it got moved the bottom of the function.

src/modules/lua_check.c
```diff
@@ -1020,6 +1021,7 @@ noit_lua_check_resume(mtev_lua_resume_info_t *ri, int nargs) {
     noit_lua_log_results(self, check);
     noit_check_end(check);
   }
+  mtev_lua_gc(ri->lmc);
```
cody271: ah matching call is here

PR #486, opened 2018-08-30
Lua leaks

postwait:


src/modules/lua_check.c
```diff
@@ -877,6 +877,7 @@ noit_lua_module_config(noit_module_t *mod,
   mtev_lua_hash_to_table(L, options);
   lua_pcall(L, 2, 1, 0);
 
+  mtevL(mtev_error, "(%s) BASE -> %d\n", mod->hdr.name, lua_gettop(L));
```
esproul: Does this need to be error, or can it be `mtev_debug` instead?

postwait: Should be removed

esproul: Done.

PR #490, opened 2018-09-11
Support account_id, check_{uuid,id} suppressions

postwait:


src/modules/fq_driver.c
```diff
@@ -236,6 +239,33 @@ static iep_thread_driver_t *noit_fq_allocate(mtev_conf_section_t conf) {
...
+      *checkid = atoi(cp);
+      mtev_hash_replace(&suppress_check_id, (const char *)checkid, sizeof(int), NULL, free, NULL);
+    }
+    free(str);
+  }
+  if(mtev_conf_get_string(conf, "suppress_check_uuid", &str)) {
+    for(cp = strtok_r(str, ",", &brk);
+        cp; cp = strtok_r(NULL, ",", &brk), i++) {
+      char *checkuuid = strdup(cp);
+      mtev_hash_replace(&suppress_check_uuid, checkuuid, sizeof(int), NULL, free, NULL);
```
rileyberton: sizeof(int) is wrong here.  checkuuid is a string, need `strlen`

PR #493, opened 2018-09-20
Support NOIT_MODULES environment variable.

postwait:
This variable contains a PCRE (default `.`) that will be
applied to each noit check module.  Module (names) that do
not match this regular expression will be skipped.

For example `NOIT_MODULES="^(selfcheck|prometheus)"` will only
load modules starting with `selfcheck` or `prometheus`.

### DIFF
--- a/src/noit_module.c
+++ b/src/noit_module.c
@@ -245,6 +245,15 @@ noit_module_help(mtev_console_closure_t ncct,
 void noit_module_init() {
   mtev_conf_section_t *sections;
   int i, cnt = 0;
+  const char *module_list = getenv("NOIT_MODULES");
+  if(!module_list) module_list = ".";
+
+  const char *error;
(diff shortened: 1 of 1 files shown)


src/noit_module.c
```diff
@@ -268,6 +277,15 @@ void noit_module_init() {
       continue;
     }
 
+    /* Check if we intend to allow this module. */
+    if(module_match) {
+      int ovector[30];
+      if(pcre_exec(module_match, NULL, module_name, strlen(module_name), 0, 0, ovector, 30) < 0) {
+        mtevL(mtev_notice, "skipping check module '%s' due to NOIT_MODULES.\n", module_name);
```
esproul: I might say "due to NOIT_MODULES environment variable." just to be crystal clear what is causing the skip.

PR #498, opened 2018-10-02
clearer error messages for REST calls

postwait:


src/noit_check_rest.c
```diff
@@ -1044,13 +1045,13 @@ rest_set_check(mtev_http_rest_closure_t *restc,
...
-    if(ocheck && ocheck != check) FAIL("new target`name would collide");
-    if(module_change) FAIL("cannot change module");
+    if(ocheck && ocheck != check) FAILC(409, "new target`name would collide");
+    if(module_change) FAILC(400, "cannot change module");
     parent = make_conf_path(pats[0]);
     if(!parent) FAIL("invalid path");
     configure_xml_check(parent, node, attr, config, &seq);
     if(check) old_seq = check->config_seq;
-    if(old_seq >= seq && seq != 0) FAIL("invalid sequence");
+    if(old_seq >= seq && seq != 0) FAILC(409, "equencing error");
```
ewilhelm: typo `"equencing `

src/noit_check_rest.c
```diff
@@ -998,15 +999,15 @@ rest_set_check(mtev_http_rest_closure_t *restc,
       if(target) xmlFree(target);
       if(name) xmlFree(name);
       if(module) xmlFree(module);
-      if(exists) FAIL("target`name already registered");
-      if(!m) FAIL("module does not exist");
+      if(exists) FAILC(409, "target`name already registered");
```
ewilhelm: Is this similar to a sequence error?  I'm not sure how this condition is triggered.

src/noit_check_rest.c
postwait: fixed

src/noit_check_rest.c
postwait: This happens if you try to instantiate two checks with the same target and name.

In circonus, all checks have names that have their circonus check_id embedded in them, so this can't happen.

PR #500, opened 2018-10-10
Fix sloppy polls_by_name maintenance.

postwait:
* target_ip was modified while node was in list.
 * target was modified while node was in list.
 * removals were skipped for sequenced checks
   (which should be done if the check will be readded)

### DIFF
--- a/src/noit_check.c
+++ b/src/noit_check.c
@@ -348,7 +348,8 @@ noit_console_show_timing_slots(mtev_console_closure_t ncct,
   return 0;
 }
 static int
-noit_check_add_to_list(noit_check_t *new_check, const char *newname) {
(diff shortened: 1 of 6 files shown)


src/noit_check.c
```diff
@@ -1611,7 +1643,7 @@ check_recycle_bin_processor(eventer_t e, int mask, void *closure,
 }
 
 int
-noit_poller_deschedule(uuid_t in, mtev_boolean log) {
+noit_poller_deschedule(uuid_t in, mtev_boolean log, mtev_boolean readding) {
```
pamaddox: This supposed to be "reading"?

postwait: heh.. no.. "re adding"  this is true if we're descheduling before we go an add it back in.

PR #501, opened 2018-10-10
XML backed *_auto_add rules are untenable.

postwait:
The XML config change rate is too high and cannot be maintained
reasonable using libxml2.  Since we're trying to manage concurrency
more than total historic unique metrics, it makes sense to only
manage these hashes transiently.  Additionally, as metrics come and
go, it makes sense to refresh the automatic set periodically. We
default this refresh to 300000ms (5 minutes) and allow it to be
set via filter_flush_period (0 to disable).

### DIFF
--- a/src/noit_filters.c
+++ b/src/noit_filters.c
@@ -59,6 +59,7 @@ static pthread_mutex_t filterset_lock;
 static pcre *fallback_no_match = NULL;
 #define LOCKFS() pthread_mutex_lock(&filterset_lock)
 #define UNLOCKFS() pthread_mutex_unlock(&filterset_lock)
+#define DEFAULT_FILTER_FLUSH_PERIOD 300000 /* 5 minutes */
 static char* filtersets_replication_path = NULL;
 
 typedef enum { NOIT_FILTER_ACCEPT, NOIT_FILTER_DENY, NOIT_FILTER_SKIPTO } noit_ruletype_t;
@@ -92,6 +93,8 @@ typedef struct _filterrule {
   char *measurement_tags;
   noit_metric_tag_search_ast_t *mtsearch;
   struct _filterrule *next;
+  struct timeval last_flush;
+  struct timeval flush_interval;
 } filterrule_t;
 
 typedef struct {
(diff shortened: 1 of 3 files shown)


src/noit_filters.c
```diff
@@ -506,18 +494,24 @@ noit_apply_filterset(const char *filterset,
...
-        if(need_name) UPDATE_FILTER_RULE(idx, name, check->name);
-        if(need_metric) UPDATE_FILTER_RULE(idx, metric, metric->metric_name);
-        noit_filterset_log_auto_add(fs->name, check, metric, r->type == NOIT_FILTER_ACCEPT);
+      if(CHECK_ADD(target) && CHECK_ADD(module) && CHECK_ADD(name) && CHECK_ADD(metric)) {
+        mtev_boolean flushed = mtev_false;
+        if(need_target) UPDATE_FILTER_RULE(target, check->target);
+        if(need_module) UPDATE_FILTER_RULE(module, check->module);
+        if(need_name) UPDATE_FILTER_RULE(name, check->name);
+        if(need_metric) UPDATE_FILTER_RULE(metric, metric->metric_name);
+        if(flushed) memcpy(&r->last_flush, &now, sizeof(now));
```
ewilhelm: How does `r->last_flush` get initialized or is `add_timeval()` okay with that?

src/noit_filters.c
```diff
@@ -435,27 +443,7 @@ noit_apply_filterrule_metric(filterrule_t *r,
...
-  }
-  child = xmlNewNode(NULL, (xmlChar *)rname);
-  xmlNodeAddContent(child, (xmlChar *)value);
-  xmlAddChild(mtev_conf_section_to_xmlnodeptr(rulenode), child);
-  CONF_DIRTY(rulenode);
-  mtev_conf_mark_changed();
-  mtev_conf_request_write();
-  mtev_conf_release_section(rulenode);
-  return 0;
-}
```
ewilhelm: This was just saving the names of metrics that were matched?  i.e. not part of how the filterset definition gets saved?

postwait: It was saving, but the saving had a bug. I fixed the bug and now saving is to slow to be useful. It also turns out that saving isn't really what we want. We want show churn.

src/noit_filters.c
postwait: It is initialized by the initial calloc of the rule (to zero)

src/noit_filters.c
```diff
@@ -59,6 +59,7 @@ static pthread_mutex_t filterset_lock;
 static pcre *fallback_no_match = NULL;
 #define LOCKFS() pthread_mutex_lock(&filterset_lock)
 #define UNLOCKFS() pthread_mutex_unlock(&filterset_lock)
+#define DEFAULT_FILTER_FLUSH_PERIOD 300000 /* 5 minutes */
```
rileyberton: name `DEFAULT_FILTER_FLUSH_PERIOD_MS` so we know the units.

src/noit_filters.c
```diff
@@ -186,7 +189,12 @@ noit_filter_compile_add(mtev_conf_section_t setinfo) {
       rule->type = (!strcmp(buffer, "accept") || !strcmp(buffer, "allow")) ?
                      NOIT_FILTER_ACCEPT : NOIT_FILTER_DENY;
     }
-
+    int32_t ffp = DEFAULT_FILTER_FLUSH_PERIOD;
+    if(!mtev_conf_get_int32(rules[j], "ancestor-or-self::node()/@filter_flush_period", &ffp))
```
rileyberton: might want to name this `filter_flush_period_ms` or use the time based config here (`300s`)

PR #503, opened 2018-10-16
If tag cat:val is too long truncate uniquely.

postwait:
This determines if the cat:val is too long and
uses a hueristic to size them appropriately and
replace with the largest possible prefix and a
suffix of _tldr_<sha1_hex> of the original cat
or val.

### DIFF
--- a/src/noit_message_decoder.c
+++ b/src/noit_message_decoder.c
@@ -400,6 +400,7 @@ noit_metric_tags_parse_one(const char *tagnm, size_t tagnmlen,
                            noit_metric_tag_t *output, mtev_boolean *toolong) {
   size_t colon_pos = 0;
   size_t cur_size = 0;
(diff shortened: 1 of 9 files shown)


src/noit_message_decoder.c
```diff
@@ -397,7 +397,7 @@ noit_metric_tags_count(const char *str, size_t strlen) {
 
 const char *
 noit_metric_tags_parse_one(const char *tagnm, size_t tagnmlen,
-                           noit_metric_tag_t *output) {
+                           noit_metric_tag_t *output, mtev_boolean *toolong) {
   size_t colon_pos = 0;
   size_t cur_size = 0;
```
rileyberton: might want to:

`*toolong = mtev_false;` 

at the top of this function due to return on line 422

PR #504, opened 2018-10-18
Send Full Metric Name With Tags To Websockets

pamaddox:
We were just using the base metric name, ignoring stream tags. These
need to be included when sending to websockets, since they are
considered part of the metric name.

### DIFF
--- a/src/noit_websocket_handler.c
+++ b/src/noit_websocket_handler.c
@@ -130,8 +130,9 @@ send_individual_metric(noit_websocket_closure_t *wcl, const char *metric_string,
 
   if (wcl->use_filter == mtev_true) {
     for (int i = 0; i < wcl->filter_count; i++) {
-      if (message.id.name_len_with_tags > 0 &&
-        strncmp(wcl->filters[i], message.id.name, message.id.name_len_with_tags) == 0) {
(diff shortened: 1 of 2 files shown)


src/noit_websocket_handler.c
```diff
@@ -130,8 +130,8 @@ send_individual_metric(noit_websocket_closure_t *wcl, const char *metric_string,
 
   if (wcl->use_filter == mtev_true) {
     for (int i = 0; i < wcl->filter_count; i++) {
-      if (message.id.name_len > 0 &&
-        strncmp(wcl->filters[i], message.id.name, message.id.name_len) == 0) {
+      if (message.id.name_len_with_tags > 0 &&
+        strncmp(wcl->filters[i], message.id.name, message.id.name_len_with_tags) == 0) {
```
postwait: This was a prexisting bug...

This strncmp isn't right we want:

`(strlen(wcl->fitlers[i]) == message.id.name_len_with_tags && 0 == memcmp(wcl->filters[i], message.id.name, message.id.name_len_with_tags))`

pamaddox: Fixed

PR #508, opened 2018-11-02
Log to */noit log_streams per default

HeinrichHartmann:
My debug logs are full of:
```
[2018-11-02 13:39:37.376822] [debug] [t@6853/e:default/7,noit_check.c:2547] q._caql`c_2438_215752::caql <- [OK]
[2018-11-02 13:39:37.376845] [debug] [t@6853/e:default/7,noit_check.c:2552] q._caql`c_2438_215752::caql -> [available:good]
[2018-11-02 13:39:37.392899] [debug] [t@6851/e:default/5,noit_check.c:2547] q._caql`c_2_198461::caql <- [OK]
[2018-11-02 13:39:37.392921] [debug] [t@6851/e:default/5,noit_check.c:2552] q._caql`c_2_198461::caql -> [available:good]
```

This change should give me a way to avoid those.

### DIFF
--- a/src/noit_mtev_bridge.c
+++ b/src/noit_mtev_bridge.c
@@ -31,11 +31,21 @@
 #include <eventer/eventer.h>
 #include <mtev_log.h>
 #include <mtev_console.h>
+#include <noit_mtev_bridge.h>
+
+mtev_log_stream_t noit_error_impl = NULL;
+mtev_log_stream_t noit_notice_impl = NULL;
+mtev_log_stream_t noit_debug_impl = NULL;
 
 static int noit_console_handler(eventer_t e, int mask, void *closure,
                                 struct timeval *now) {
   return mtev_console_handler(e,mask,closure,now);
 }
+
 void noit_mtev_bridge_init() {
+  noit_error_impl = mtev_log_stream_find("error/noit");
+  noit_notice_impl = mtev_log_stream_find("notice/noit");
+  noit_debug_impl = mtev_log_stream_find("debug/noit");
+
   eventer_name_callback("noit_console", noit_console_handler);
 }
--- a/src/noit_mtev_bridge.h
+++ b/src/noit_mtev_bridge.h
@@ -33,9 +33,13 @@
 
 #include <mtev_log.h>
 
-#define noit_debug mtev_debug
-#define noit_error mtev_error
-#define noit_notice mtev_notice
+extern mtev_log_stream_t noit_error_impl;
+extern mtev_log_stream_t noit_notice_impl;
+extern mtev_log_stream_t noit_debug_impl;
+
+#define noit_error ( (noit_error_impl) ? (noit_error_impl) : (mtev_error) )
+#define noit_notice ( (noit_notice_impl) ? (noit_notice_impl) : (mtev_notice) )
+#define noit_debug ( (noit_debug_impl) ? (noit_debug_impl) : (mtev_debug) )
 #define noit_stderr mtev_stderr
 
 API_EXPORT(void)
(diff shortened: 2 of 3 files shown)


src/noit_mtev_bridge.h
```diff
@@ -33,9 +33,13 @@
 
 #include <mtev_log.h>
 
-#define noit_debug mtev_debug
-#define noit_error mtev_error
-#define noit_notice mtev_notice
+mtev_log_stream_t noit_error_impl;
+mtev_log_stream_t noit_notice_impl;
+mtev_log_stream_t noit_debug_impl;
```
postwait: These must be `extern` in a header and the actual declarations in a C file somewhere.  This is where it gets more complicated b/c of libnoit.

HeinrichHartmann: Updated. Now running into this error:

```
noit_check_log_helpers.o: In function `noit_check_log_bundle_compress_b64':
//home/vagrant/circ/src/reconnoiter/src/noit_check_log_helpers.c:101: undefined reference to `noit_error_impl'
//home/vagrant/circ/src/reconnoiter/src/noit_check_log_helpers.c:101: undefined reference to `noit_error_impl'
//home/vagrant/circ/src/reconnoiter/src/noit_check_log_helpers.c:101: undefined reference to `noit_error_impl'
//home/vagrant/circ/src/reconnoiter/src/noit_check_log_helpers.c:101: undefined reference to `noit_error_impl'
//home/vagrant/circ/src/reconnoiter/src/noit_check_log_helpers.c:101: undefined reference to `noit_error_impl'
noit_check_log_helpers.o://home/vagrant/circ/src/reconnoiter/src/noit_check_log_helpers.c:101: more undefined references to `noit_error_impl' follow
```

HeinrichHartmann: noit_check_log_helpers.c does include noit_mtev_bridge.h, but apparently the linker can not resolve the noit_error_impl symbol. Do we need to add a linker flag here `-Lsomething` ?

HeinrichHartmann: Tweaked the Makefile a little bit. Seems to build now on my box.

postwait: The tricky part here is that those symbols need to be in one object and not in more than one.  We have the library and the binaries (noitd, stratcond, noit_b2sm).  This will require a trick to pull off.

there's also the question in that if I use libnoit (from say snowthd), should it be logging to these "noit/debug" or should it be to general debug?

HeinrichHartmann: I certainly want libnoit to log into substreams `error/noit`, etc., so that I can keep an eye on `error` and selectively disable noit logs when I am not interested in them.

Have you seen the changes to the makefile? Looks like this is working as intended. I had a few double definition errors but got them resolved.

src/noit_mtev_bridge.c
```diff
@@ -31,11 +31,21 @@
 #include <eventer/eventer.h>
 #include <mtev_log.h>
 #include <mtev_console.h>
+#include <noit_mtev_bridge.h>
+
+mtev_log_stream_t noit_error_impl;
```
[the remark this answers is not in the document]
HeinrichHartmann: Done. But I am confused: Are these variables not automatically initialized to NULL (since they are defined outside of a function?) ?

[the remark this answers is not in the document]
postwait: They are by the C spec.  However, as they *may* be initialized to something other than zero and stack variables are not initialized to zero we maintain the convention to be explicit always.

PR #512, opened 2018-12-18
metric_director extensions for lua tag-search

HeinrichHartmann:
caql-broker tag-search PR: https://github.com/circonus/caql-broker/pull/81

Tests are in the caql-broker repository, since there is no test "harness" for lua in here. (TODO: Move tests here)

### DIFF
--- a/src/modules/noit_lua_libnoit_binding.c
+++ b/src/modules/noit_lua_libnoit_binding.c
@@ -297,7 +297,7 @@ noit_lua_tagset_copy_free(lua_State *L) {
 static int
 noit_lua_tagset_copy_setup(lua_State *L, noit_metric_tagset_t *src_set) {
   noit_metric_tagset_t *dst_set = calloc(1, sizeof(*src_set));
-  memcpy(dst_set, src_set, sizeof(dst_set));
+  memcpy(dst_set, src_set, sizeof(*dst_set));
   noit_metric_tag_t *dst_tags = calloc(src_set->tag_count, sizeof(noit_metric_tag_t));
   memcpy(dst_tags, src_set->tags, src_set->tag_count*sizeof(noit_metric_tag_t));
   dst_set->tags = dst_tags;
--- a/src/modules/noit_lua_libnoit_binding.c
+++ b/src/modules/noit_lua_libnoit_binding.c
@@ -71,6 +71,22 @@ lua_noit_checks_unsubscribe(lua_State *L) {
   return lua_noit_checks_adjustsubscribe(L, -1);
 }
 
+mtev_hook_return_t
+hook_metric_subscribe_all(void *closure, noit_metric_message_t *m, int *w, int wlen) {
+  int *lane = (int*) closure;
+  assert(*lane < wlen);
+  w[*lane] = 1;
+  return MTEV_HOOK_CONTINUE;
+}
+
+static int
+lua_noit_metric_subscribe_all(lua_State *L) {
+  int *lane = malloc(sizeof(int));
+  *lane = noit_metric_director_my_lane();
+  metric_director_want_hook_register("metrics_select", hook_metric_subscribe_all, lane);
+  return 0;
+}
+
 static int
 noit_lua_tagset_copy_setup(lua_State *, noit_metric_tagset_t *);
 
@@ -406,6 +422,7 @@ static const luaL_Reg libnoit_binding[] = {
   { "metric_director_next", lua_noit_metric_next },
   { "metric_director_get_messages_received", lua_noit_metric_messages_received },
   { "metric_director_get_messages_distributed", lua_noit_metric_messages_distributed},
+  { "metric_director_subscribe_all", lua_noit_metric_subscribe_all},
   { "tag_parse", lua_noit_tag_parse},
   { "tag_tostring", lua_noit_tag_tostring},
   { "tag_search_parse", lua_noit_tag_search_parse},
--- a/src/modules/noit_lua_libnoit_binding.c
+++ b/src/modules/noit_lua_libnoit_binding.c
@@ -36,7 +36,16 @@
 #include "lua.h"
 #include "noit_metric_tag_search.h"
 
+// We need to hold-on to a reference to the metric name as long as we make use of the tagset
+typedef struct {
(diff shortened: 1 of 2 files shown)


src/modules/noit_lua_libnoit_binding.c
```diff
@@ -246,6 +322,149 @@ lua_noit_metric_messages_distributed(lua_State *L) {
...
+    luaL_error(L, "Error parsing tagset for metric %s", name);
+    return 0;
+  }
+  mtev_boolean ok = noit_metric_tag_search_evaluate_against_tags(*ast_ud, &mtset);
+  lua_pushboolean(L, ok);
+  return 1;
+}
+
+void
+libnoit_init() {
```
HeinrichHartmann: Call this function explicitly if you want to make use of this somewhere.

HeinrichHartmann: Add locks, so we are not getting multiple copies of account_set

src/modules/noit_lua_libnoit_binding.c
```diff
@@ -64,11 +68,71 @@ static int
...
+  // TODO: Call this during startup. Where to put this?
+  if(account_set == NULL) {
+    libnoit_init();
+  }
+  int account_id = luaL_checkint(L, 1);
+  int *account_interests = NULL;
+  if(!mtev_hash_retrieve(account_set, (const char *)&account_id, sizeof(account_id), (void **)&account_interests)) {
+    int nthreads = eventer_loop_concurrency();
+    account_interests = calloc(nthreads, (sizeof(*account_interests)));
+    int rc = mtev_hash_store(account_set, (const char *)&account_id, sizeof(account_id), (void **)account_interests);
```
postwait: This address is on stack... it will not work.  You need a copy of this:
```
int *account_id_copy = calloc(1, sizeof(int));
*account_id_copy = account_id;
mtev_hash_store(account_set, account_id_copy, sizeof(int), (void **)account_interests);
```

src/modules/noit_lua_libnoit_binding.c
```diff
@@ -246,6 +322,149 @@ lua_noit_metric_messages_distributed(lua_State *L) {
...
+    return 0;
+  }
+  mtev_boolean ok = noit_metric_tag_search_evaluate_against_tags(*ast_ud, &mtset);
+  lua_pushboolean(L, ok);
+  return 1;
+}
+
+void
+libnoit_init() {
+  account_set = calloc(1, sizeof(*account_set));
```
postwait: Locks required and initialization before set using ck_pr_ptr_store

src/modules/noit_lua_libnoit_binding.c
```diff
@@ -246,6 +322,149 @@ lua_noit_metric_messages_distributed(lua_State *L) {
...
+  }
+  return cnt;
+}
+
+// Parse a metric name to a tagset
+// param: name
+// returns: stags, mtags (userdata)
+static int
+lua_noit_tag_parse(lua_State *L) {
+  const char* name = luaL_checkstring(L, 1);
```
HeinrichHartmann: Make sure this name string lives as long as the tagsets being used in lua. This can be done be lua-refing the name from the mtset and the stset userdata.

src/modules/noit_lua_libnoit_binding.c
```diff
@@ -246,6 +322,149 @@ lua_noit_metric_messages_distributed(lua_State *L) {
...
+  noit_metric_tag_t stags[MAX_TAGS], mtags[MAX_TAGS];
+  noit_metric_tagset_t stset = { .tags = stags, .tag_count = MAX_TAGS };
+  noit_metric_tagset_t mtset = { .tags = mtags, .tag_count = MAX_TAGS };
+  int rc = noit_metric_parse_tags(name, strlen(name), &stset, &mtset);
+  if (rc < 0) {
+    luaL_error(L, "Error parsing tags for metric %s", name);
+    return 0;
+  }
+  noit_lua_tagset_copy_setup(L, &stset);
```
postwait: pass in a lua_ref to the name lua_ref(L,1) and store it in the udata for the lua tagset, unrefing it on __gc.

src/modules/noit_lua_libnoit_binding.c
```diff
@@ -64,11 +68,71 @@ static int
...
+
+static int
+lua_noit_metric_subscribe_account(lua_State *L) {
+  // TODO: Call this during startup. Where to put this?
+  if(account_set == NULL) {
+    libnoit_init();
+  }
+  int account_id = luaL_checkint(L, 1);
+  int *account_interests = NULL;
+  if(!mtev_hash_retrieve(account_set, (const char *)&account_id, sizeof(account_id), (void **)&account_interests)) {
```
HeinrichHartmann: wrap that into a while() function, so we get thread safe.

src/modules/noit_lua_libnoit_binding.c
```diff
@@ -246,6 +341,169 @@ lua_noit_metric_messages_distributed(lua_State *L) {
...
+  // Using a spinlock instead, since it does not require initialization
+  ck_spinlock_lock(&noit_lua_libnoit_init_lock);
+  if (account_set == NULL) {
+    mtev_hash_table *tmp = calloc(1, sizeof(*account_set));
+    mtev_hash_init(tmp);
+    ck_pr_store_ptr(&account_set, tmp);
+    metric_director_want_hook_register("metrics_select", hook_metric_subscribe_account, NULL);
+  }
+  ck_spinlock_unlock(&noit_lua_libnoit_init_lock);
+}
```
HeinrichHartmann: This is the best I could come up with to implement this logic suggested by Theo in chat:
```
pthread_mutex_lock();
if(account_set == NULL) {
  mtec_hash_table *tmptable  = calloc....
  mtec_hash_init(tmptable);
  ck_pr_ptr_store(&account_set, tmptable);
  hook_register...
}
pthead_mutex_unlock();
```
I could not find a good place to initialize the pthread lock, nor where to call this function from during initialization phase. Found the spinlock in CK which can be initialized with a constant.
Not sure if this is a sane solution, but it appears to work 🤞.

PR #515, opened 2019-03-01
upd: max metric name length limit to MAX_METRIC_TAGGED_NAME

maier:
httptrap is rejecting PUTs when one or more metrics have names > 255 characters. This happens frequently when stream tags are added to metric names. Using MAX_METRIC_TAGGED_NAME from noit_metric.h. Unclear if the 255 limit set in httptrap is intentional due to a constraint somewhere else in the metric handling path.

### DIFF
--- a/src/modules/httptrap.c
+++ b/src/modules/httptrap.c
@@ -6,7 +6,7 @@
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are
  * met:
- * 
+ *
  *     * Redistributions of source code must retain the above copyright
  *       notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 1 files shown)


bdunavant:
While reviewing the rest of the file for like issues, I noticed this on line 878.
`json_object_object_add(value_obj, "_value", json_object_new_string(buff));`
buff is 256, so if debug is on, we would output a truncated version of any text metrics that are longer than 255 chars.  I think this is probably safe but thought I'd note it.

maier:
I believe that is for the type and value, not the name. I may be misreading it though.

postwait:
Brian's right about that 256 buffer.

PR #518, opened 2019-03-14
Fix and test filter stream_tags.

postwait:
* metric matching only matches against the metric base name (not tags).
 * __name is inserted and made available to the stream_tags expression.

### DIFF
--- a/src/noit_check_rest.c
+++ b/src/noit_check_rest.c
@@ -79,7 +79,7 @@ static int
 rest_show_config(mtev_http_rest_closure_t *, int, char **);
 
 static void
-add_metrics_to_node(stats_t *c, xmlNodePtr metrics, const char *type,
+add_metrics_to_node(noit_check_t *check, stats_t *c, xmlNodePtr metrics, const char *type,
                     int include_time, mtev_hash_table *supp) {
   mtev_hash_table *mets;
   mtev_hash_iter iter = MTEV_HASH_ITER_ZERO;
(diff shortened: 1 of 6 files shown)


src/noit_filters.c
```diff
@@ -482,7 +495,7 @@ noit_apply_filterset(const char *filterset,
       need_target = !MATCHES(target, check->target);
       need_module = !MATCHES(module, check->module);
       need_name = !MATCHES(name, check->name);
-      need_metric = !noit_apply_filterrule_metric(r, metric->metric_name, strlen(metric->metric_name), &stset, &mtset);
+      need_metric = !noit_apply_filterrule_metric(r, metric->metric_name, mlen, &stset, &mtset);
```
ewilhelm: IIUC, this is taking the mlen from noit_metric_parse_tags(), which is the length of the metric name without tags?

bdunavant: noit_metric_parse_tags() returns the input len passed into it, which is `strlen(metric->metric_name)`.  It's the same value, we just aren't recomputing it a second time.

ewilhelm: ...and noit_apply_filterrule_metric() already supported stream tags `stsearch`

src/noit_check_rest.c
```diff
@@ -101,6 +101,9 @@ add_metrics_to_node(stats_t *c, xmlNodePtr metrics, const char *type,
     xmlSetProp(tmp, (xmlChar *)"name", (xmlChar *)m->metric_name);
     buff[0] = m->metric_type; buff[1] = '\0';
     xmlSetProp(tmp, (xmlChar *)"type", (xmlChar *)buff);
+    if(!noit_apply_filterset(check->filterset, check, m)) {
+      xmlSetProp(tmp, (xmlChar *)"filtered", (xmlChar *)"true");
+    }
```
ewilhelm: I don't see any comment about this new filtered / _filtered attribute in the xml or json

src/noit_filters.c
postwait: noit_metric_parse_tags returns the length without andy tags.  The "eat_up_tags" function in there modifies (down) input_len as it ... eats up tags.

bdunavant: Ah, i missed the `eat_up_tags(input, &input_len`

postwait:
addressed

src/noit_check_rest.c
postwait: I noticed that the check pulling has no way to show which metrics are currently filtered out.  so I added an attribute.  I didn't add the "false" case b/c the documents are large and didn't need to get much larger.

PR #520, opened 2019-04-05
Fix Capitalization Of Accept Header

pamaddox:
This is case sensitive - lower-case 'a' "accept" won't work. Needs to be
capitalized.


pamaddox:
need to fix in library itself, not here

PR #525, opened 2019-05-07
 Remove obsolete rejection of tags with special chars, this is handled in canonicalize

vasuraman11:


vasuraman11:
note that although this branch is based on ci-opentsdb_ingestion, it will not be stacked but will be rebased to master after the other branch is merged (I won't repeat our earlier issue with GitHub being confused with stacked branches that were reverted)

PR #526, opened 2019-05-09
CIRC-3596 Add some mtev logging for httptrap on errors or httptrap debug header

vasuraman11:
This will help debug issues with timeouts at MLB, could also help with cause of 500's.

### DIFF
--- a/src/modules/httptrap.c
+++ b/src/modules/httptrap.c
@@ -822,7 +822,7 @@ rest_httptrap_handler(mtev_http_rest_closure_t *restc,
   }
 
   if (debugflag) {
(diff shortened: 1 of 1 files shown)


src/modules/httptrap.c
```diff
@@ -851,6 +855,8 @@ rest_httptrap_handler(mtev_http_rest_closure_t *restc,
       json_object *metrics_obj;
       metrics_obj = json_object_new_object();
 
+      mtevL(nldeb, "HTTPTRAP: Processed %d records\n", cnt);
```
postwait: The nldeb logging facility is already named debug/httptrap so the extra HTTPTRAP is redundant.

vasuraman11: done

PR #528, opened 2019-05-17
Bundled immediate trap

postwait:


src/noit_check_log.c
```diff
@@ -548,35 +597,51 @@ noit_check_log_bundle_fb_serialize(mtev_log_stream_t ls, noit_check_t *check) {
...
 
-  size_t fb_size;
-  void *buffer = noit_fb_finalize_metricbatch(B, &fb_size);
-  char *outbuf;
-  unsigned int outsize;
-  noit_check_log_bundle_compress_b64(NOIT_COMPRESS_LZ4, buffer, fb_size, &outbuf, &outsize);
+    if(current_in_batch >= metrics_per_bundle) {
+do_batch:
+      buffer = noit_fb_finalize_metricbatch(B, &fb_size);
+      noit_check_log_bundle_compress_b64(NOIT_COMPRESS_LZ4, buffer, fb_size, &outbuf, &outsize);
```
vasuraman11: outbuf can come back NULL when this method returns -1, in which case the mtevL a couple lines down will segfault

src/noit_check_log.c
```diff
@@ -684,6 +752,8 @@ noit_check_log_bundle_serialize(mtev_log_stream_t ls, noit_check_t *check) {
     // Compress + B64
     comp = use_compression ? NOIT_COMPRESS_ZLIB : NOIT_COMPRESS_NONE;
     noit_check_log_bundle_compress_b64(comp, buf, size, &out_buf, &out_size);
+    mtevL(mtev_debug, "B%c compression: %f%%\n", use_compression ? '1' : '2',
```
vasuraman11: out_size could be undefined (never initialized), there are three return paths in noit_check_log_bundle_compress_b64() that don't set out_buf or out_size at all.

PR #535, opened 2019-06-18
Use 4xx codes for httptrap errors

postwait:


src/modules/httptrap.c
```diff
@@ -888,6 +905,7 @@ rest_httptrap_handler(mtev_http_rest_closure_t *restc,
 
     if(!allowed) {
       error = "secret mismatch";
+      error_code = 403;
```
bdunavant: Because the secret is part of the URL, I think this should technically also be a 404, rather than a 403 (which implies a valid URL) but whatever.

postwait: HTTP codes map poorly... arguably (to the user) the password is part of a URL like http://user:pass@host:port/....

The goal is to inform a user when we can confirm their identity, but not their access.

PR #538, opened 2019-06-26
Allow dropping of late arriving data in metric_director

HeinrichHartmann:


src/noit_metric_director.c
```diff
@@ -436,6 +437,13 @@ handle_metric_buffer(const char *payload, int payload_len,
 
         int rv = noit_message_decoder_parse_line(message, has_noit);
 
+        uint64_t drop_before_ms;
+        if((drop_before_ms = ck_pr_load_64(&drop_before_threshold_ms)) > 0) {
+          if(message->value.whence_ms < drop_before_ms) {
+            goto bail;
+          }
+        }
+
```
HeinrichHartmann: Comments:
1) We could have done the check before parsing the message using `get_message_time`. This would have added additional parsing overhead for the common case (?) that the message arrives in time. Here we just have to pay for a load + compare. On the flip-side the case of late arriving data involves calloc+free of a metric_message_t and parsing of the full message. Not ideal, but much better than forwarding that stuff all the way to lua.

2) Not sure the ck_pr_load_64 is necessary here. Maybe we could have just worked with the global variable instead of making a stack copy?! At least in my use case the drop_before_threshold only ever increases. So stale reads do cause a few messages more to pass through.

postwait: Ck stuff is unnecessary here.

postwait:
We already forego sending old messages to fq from within stratcon. Do think that isn't working?

HeinrichHartmann:
Discussed elsewhere:
- Yes, we see late arriving data in practice.
- This is a threshold not a difference to WCT. So it changes frequently, and is not suitable for a config option.

src/noit_metric_director.c
```diff
@@ -577,3 +586,7 @@ noit_metric_director_get_messages_distributed() {
  return ck_pr_load_64(&number_of_messages_distributed);
 }
 
+void
+noit_metric_director_drop_before(double t) {
+  drop_before_threshold_ms = (uint64_t) t*1000;
```
postwait: You want `(uint64_t) (t*1000.0)`

As is, t is converted to a uint64_t before multiplying and will always look milliseconds.

HeinrichHartmann: Removed the cast.

PR #539, opened 2019-06-26
CIRC-3566 Ping_icmp module test check fixes

vasuraman11:
PIng_icmp currently fails to match up ping responses with a test check, so running a test check always fails.  This was due to a tightening up of code done 8 months ago where code was added to make sure that the ping is coming from a valid check being polled (but test checks are not in the polling list because they are temporary).
To fix this we add a new hash and access function in the testcheck module which allows another check module to see if the check is a test check.  This function is dynamically linked and will do nothing if the testcheck module is not loaded.  If ping_icmp cannot find the check by doing noit_poller_lookup(), then it will fallback to looking for a test check (if that module is loaded).

### DIFF
--- a/src/modules/check_test.c
+++ b/src/modules/check_test.c
@@ -40,11 +40,14 @@
 #include "noit_check.h"
 #include "noit_check_rest.h"
 
+#include "check_test.h"
 #include "check_test.xmlh"
 
 static void check_test_schedule_sweeper();
 static mtev_log_stream_t nlerr = NULL;
 static mtev_log_stream_t nldeb = NULL;
+static mtev_hash_table test_checks;
+static pthread_mutex_t test_checks_lock = PTHREAD_MUTEX_INITIALIZER;
 
 struct check_test_closure {
   noit_check_t *check;
@@ -176,6 +179,8 @@ noit_fire_check(xmlNodePtr attr, xmlNodePtr config, const char **error) {
   c->flags |= NP_DISABLED; /* this is hack to know we haven't run it yet */
   if(NOIT_CHECK_SHOULD_RESOLVE(c))
     noit_check_resolve(c);
+  mtev_hash_replace(&test_checks, (char *)c->checkid, UUID_SIZE, c, NULL,
+                    (NoitHashFreeFunc)noit_poller_free_check);
 
  error:
   if(conf_hash) {
@@ -246,7 +251,8 @@ rest_test_check_result(struct check_test_closure *cl) {
     }
   }
 
-  noit_poller_free_check(cl->check);
+  // this should free the check automatically with noit_poller_free_check
+  mtev_hash_delete(&test_checks, (char *)cl->check->checkid, UUID_SIZE, NULL, NULL);
   free(cl);
 }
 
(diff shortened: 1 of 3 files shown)


src/modules/check_test.h
```diff
@@ -0,0 +1,27 @@
+/* Copyright 2019, Circonus Inc.
+ * All Rights Reserved.
```
esproul: Reconnoiter is BSD-licensed. New headers should carry some version of the license, probably just with a Circonus copyright.

vasuraman11: Good point - fixed with same copyright header we are using on other reconnoiter source,

src/modules/check_test.h
```diff
@@ -0,0 +1,54 @@
...
+                     (in))
+
+#define NOIT_TESTCHECK_LOOKUP(noit_check_ptr,uuid) do { \
+  if(noit_testcheck_lookup_available()) { \
+    noit_check_ptr = noit_testcheck_lookup(uuid); \
+  } \
+  else { \
+    noit_check_ptr = NULL; \
+  } \
+} while(0)
```
postwait: I am not a fan of global macros that treat args are lvalues -- it's very un C.

```
#define NOIT_TESTCHECK_LOOKUP(uuid) \
  (noit_testcheck_lookup_available() ? noit_testcheck_lookup(uuid) : (noit_check_t *)NULL)
```

src/modules/ping_icmp.c
```diff
@@ -322,12 +338,31 @@ static int ping_icmp_handler(eventer_t e, int mask,
...
-    if(payload->check_pack_cnt != data->expected_count) continue;
-    if(payload->check_pack_no >= data->expected_count) continue;
+    if(payload->check_no != data->check_no) {
+      mtevLT(nldeb, now,
+             "ping_icmp response check number mismatch for check '%s'\n", uuid_str);
+      continue;
+    }
+    if(payload->check_pack_cnt != data->expected_count) {
+      mtevLT(nldeb, now,
+             "ping_icmp response check pack count mismatch for check '%s'\n", uuid_str);
```
esproul: Small nit: can we just say "packet" in the log message? I assume that is what the structure member refers to.  Same for line 360.

vasuraman11:
you're right, but why is `noit_poller_lookup()` then locking in `noit_check.c`?

src/modules/check_test.h
vasuraman11: will commit this, sounds good

src/modules/ping_icmp.c
vasuraman11: sure

src/modules/check_test.h
vasuraman11: fixed

vasuraman11:
Review comments addressed, switched hash init to set it up to be thread-safe without needing pthread locks

src/modules/check_test.c
```diff
@@ -365,13 +370,22 @@ rest_test_check(mtev_http_rest_closure_t *restc,
 
 static int
 check_test_init(mtev_dso_generic_t *self) {
+  mtev_hash_init_locks(&test_checks, 32, MTEV_HASH_LOCK_MODE_MUTEX);
   mtevAssert(mtev_http_rest_register(
-    "POST", "/checks/", "^test(\\.xml|\\.json)?$",
-    rest_test_check
-  ) == 0);
+                 "POST", "/checks/", "^test(\\.xml|\\.json)?$",
```
vasuraman11: this one is odd, going to fix - must be some editor "helpfulness"

PR #542, opened 2019-06-27
Support filering of metrics with tags using canonical names.

postwait:
Previously <metric>foo</metric> filtersets would match only
against the untagged portion of the name.  This change will allo
<metric>foo|ST[a:b]</metric> match exactly that canonical name.

Improve filter test to test new metric name matching, more
thorough evaluation of existing test, and auto_add hash filters
including reset.

### DIFF
--- a/src/noit_filters.c
+++ b/src/noit_filters.c
@@ -208,7 +208,7 @@ noit_filter_compile_add(mtev_conf_section_t setinfo) {
       rule->ruleid = strdup(buffer);
     }
     /* Compile any hash tables, should they exist */
-#define HT_COMPILE(rname) do { \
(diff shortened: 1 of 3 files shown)


ewilhelm:
If I'm following correctly, this changes the matching for `metric=` to include the full and canonical streamtags, but what about regular expression matching?  Does an end-of-string anchor now mean the end of the streamtags component?

postwait:
pattern should never match into the tags, that would be broken by design as they are encoded and comma separated.  Basically... patterns and name matching go against the name of the metric (without tags).  The change is that in the special case that a name (not pattern) is specified in canonical form that it will attemt to match that against a canonically formed metric name.

PR #547, opened 2019-07-19
Don't convert wildcard expansions in Base64 encoded tags

cody271:


postwait:
This means that if you want wildcard expansion and you can't express it without base64 encoding, you have actually form a regex using b// approach.  That should be noted.  It will break the current Circonus stuff.

cody271:
Replaced with PR #549.

PR #555, opened 2019-08-21
Fix clustered histogram submission with explicit _ts

postwait:
Histogram submission is intentionally jittered when running in
a clustered configuration. However, when _ts is specified, that
jitter should not be applied. This change makes it such that
jitter is not applied when logging a histogram with an explicit
_ts setting.

### DIFF
--- a/src/modules/histogram.c
+++ b/src/modules/histogram.c
@@ -161,8 +161,8 @@ noit_log_histo_encoded_function_validate(noit_check_t *check, struct timeval *wh
   }
   mtev_uuid_unparse_lower(check->checkid, uuid_str + strlen(uuid_str));
 
-  unsigned long ms_cluster_jitter = noit_cluster_self_index() + 1;
-  if(explicit_time) ms_cluster_jitter = 0;
+  unsigned long ms_cluster_jitter = 0
+  if(!explicit_time) ms_cluster_jitter = noit_cluster_self_index() + 1;
 
 #define SECPART(a) ((unsigned long)(a)->tv_sec)
 #define MSECPART(a) ((unsigned long)((a)->tv_usec / 1000) + ms_cluster_jitter)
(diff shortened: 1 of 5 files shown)


src/modules/histogram.c
```diff
@@ -162,6 +162,7 @@ noit_log_histo_encoded_function_validate(noit_check_t *check, struct timeval *wh
   mtev_uuid_unparse_lower(check->checkid, uuid_str + strlen(uuid_str));
 
   unsigned long ms_cluster_jitter = noit_cluster_self_index() + 1;
+  if(explicit_time) ms_cluster_jitter = 0;
```
bdunavant: We should probably initialize ms_cluster_jitter to 0, and then `if(!explicit_time) ms_cluster_jitter = noit_cluster_self_index() + 1;` to avoid making that call.   For a non-clustered broker, noit_cluster_self_index() is going to call mtev_cluster_by_name() every iteration, so I think it would be slightly faster in this case.

src/modules/httptrap.c
```diff
@@ -418,6 +417,15 @@ httptrap_yajl_cb_string(void *ctx, const unsigned char * stringVal,
     json->saw_complex_type |= HT_EX_VALUE;
     return 1;
   }
+  else if(json->last_special_key == HT_EX_TS) {
+    uint64_t ts = strtoull(stringVal, NULL, 10);
```
bdunavant: Do we care if there is garbage there?  Should we be checking if this returns 0, and returning 0 as well, rather than going forward with a 0 timestamp?

postwait: All arguable.  A _ts of 0 will eventually get treated correctly, so that seems like a non-issue.

I'm fine making this strict that the string be completely parsable, it's a change from elsewhere though.

PR #557, opened 2019-08-23
Optimize for reading small chunks in a stream.

postwait:
This reduces reading from a hostile endpoint sending small chunks
(10-20bytes) for a 1Mb payload from 25s down to 0.5s.

### DIFF
--- a/src/modules-lua/noit/module/resmon.lua
+++ b/src/modules-lua/noit/module/resmon.lua
@@ -280,12 +280,12 @@ function initiate(module, check)
         error(schema .. " not supported")
     end 
 
-    local output = ''
+    local output_tbl = {''}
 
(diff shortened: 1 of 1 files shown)


src/modules-lua/noit/module/resmon.lua
```diff
@@ -383,26 +385,31 @@ function initiate(module, check)
...
 
+    local output = table.concat(output_tbl, "")
+
     if helper and helper.fix_output then output = helper.fix_output(output) end
 
     if helper and helper.process then
       helper.process(check, output, client.headers)
       return
     end
+mtev.log("error", "post-processed -> %s\n", uri)
```
bdunavant: whoops

PR #574, opened 2020-02-07
Support legacy statsd concepts.

postwait:
legacy=true in the check config will avoid appending statsd type tags
and use regular numeric stats for |c count metrics instead of
histograms.  This concession does not have an elegant clustering story.
Timing metrics are still treated as histograms as the alternative is an
assualt on dignity.


postwait:
This fundamentally breaks clustering guarantees and is just not worth supporting.

PR #575, opened 2020-02-14
Execute tag searches if metric is absent.

postwait:


cody271:
Is it possible to add a test for this easily? I see `busted/simple/filters/filter_spec.lua`..

PR #578, opened 2020-02-28
Graphite pickle

postwait:


docs/config/modules/graphite_pickle.xml
```diff
@@ -0,0 +1,105 @@
+<?xml version="1.0"?>
+<section xmlns="http://docbook.org/ns/docbook" version="5">
+  <title>graphite_pickle</title>
+  <para>The graphite module provides a simple way to push data into reconnoiter via the carbon-relay pickle format.  Simply configure the relay to write to the provided port.</para>
```
vasuraman11:
```suggestion
  <para>The graphite pickle module provides a simple way to push data into reconnoiter via the carbon-relay pickle format.  Simply configure the relay to write to the provided port.</para>
```
```suggestion
  <para>The graphite module provides a simple way to push data into reconnoiter via the carbon-relay pickle format.  Simply configure the relay to write to the provided port.</para>
```

vasuraman11: disregard the no change suggestion, accidental double mouse click

vasuraman11: actually it is the graphite module, so you can disregard this, just wanted to make it clear to user.

PR #590, opened 2020-03-13
Implement statsd_tcp

postwait:


src/modules/statsd.c
```diff
@@ -492,3 +494,185 @@ noit_module_t statsd = {
...
+        mtevL(noit_error,
+              "statsd: could not set socket (IPv4) non-blocking: %s\n",
+              strerror(errno));
+        return -1;
+      }
+      socklen_t reuse = 1;
+      if (setsockopt(ccl->ipv4_listen_fd, SOL_SOCKET, SO_REUSEADDR, (void *)&reuse, sizeof(reuse)) != 0) {
+        mtevL(noit_error, "statsd listener(IPv4) failed(%s) to set REUSEADDR (doing our best)\n", strerror(errno));
+      }
+#ifdef SO_REUSEPORT
```
bdunavant: If SO_REUSEPORT is defined, we should run the above code, AND the below code?  e.g. we call setsockops() twice?   This could be correct, but just looks weird to me.

postwait: one is REUSEADDR (which is everywhere) the other is REUSEPORT (which is protected due to limited availability)

PR #597, opened 2020-04-02
CIRC-4638 - Add LMDB Backing Option For Checks

pamaddox:
Added a flag to the "checks" configuration - "use_lmdb" - that, if set to true,
will enable an LMDB db for storing checks instead of XML. This will require setting
an "lmdb_path" field as well to tell it where the LMDB database will be stored. In
addition, there is a true/false flag, "convert_xml_to_lmdb", that will convert the existing
XML backingstore to use LMDB.

LMDB reads/writes are considerably faster than XML configuration - this path will enable
a significant speed-up when reading/writing checks - particularly noticeable if there are
a considerable number of checks.

### DIFF
--- a/src/noit_check_lmdb.c
+++ b/src/noit_check_lmdb.c
@@ -1282,23 +1282,6 @@ noit_check_lmdb_process_repl(xmlDocPtr doc) {
       continue;
     }
 
-    if(check) {
-#if 0
-      char xpath[1024];
-
-      snprintf(xpath, sizeof(xpath), "/noit/checks//check[@uuid=\"%s\"]",
-               uuid_str);
-      mtev_conf_section_t oldsection = mtev_conf_get_section_write(MTEV_CONF_ROOT, xpath);
-      if(!mtev_conf_section_is_empty(oldsection)) {
-        CONF_REMOVE(oldsection);
-        node = mtev_conf_section_to_xmlnodeptr(oldsection);
-        xmlUnlinkNode(node);
-        xmlFreeNode(node);
-      }
-      mtev_conf_release_section_write(oldsection);
-#endif
-    }
-
 #if 0
     xmlNodePtr checks_node = mtev_conf_section_to_xmlnodeptr(checks);
     child = xmlDocCopyNode(child, checks_node->doc, 1);
@@ -1307,11 +1290,6 @@ noit_check_lmdb_process_repl(xmlDocPtr doc) {
     noit_poller_process_check_conf(section);
 #endif
     i++;
-  }
-  mtev_conf_release_section_write(checks);
-  mtev_conf_mark_changed();
-  if(mtev_conf_write_file(NULL) != 0) {
-    mtevL(mtev_error, "local config write failed\n");
   }
   return i;
 }
--- a/src/noit_check_lmdb.c
+++ b/src/noit_check_lmdb.c
@@ -1163,10 +1163,7 @@ noit_check_lmdb_convert_one_xml_check_to_lmdb(mtev_conf_section_t section, char 
   }
 
   if (!MYATTR(stringbuf, resolve_rtype, resolve_rtype, sizeof(resolve_rtype))) {
-    if (!INHERIT(stringbuf, resolve_rtype, resolve_rtype, sizeof(resolve_rtype))) {
-      strlcpy(resolve_rtype, PREFER_IPV4, sizeof(resolve_rtype));
-    }
-    else {
+    if (INHERIT(stringbuf, resolve_rtype, resolve_rtype, sizeof(resolve_rtype))) {
       resolve_rtype_inherited = mtev_true;
     }
   }
--- a/src/noit_check_lmdb.c
+++ b/src/noit_check_lmdb.c
@@ -530,6 +530,7 @@ put_retry:
   }
   mdb_cursor_close(cursor);
   mtev_hash_destroy(&conf_table, free, NULL);
+  ck_rwlock_read_unlock(&instance->lock);
 }
 
 int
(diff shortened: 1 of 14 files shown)


src/noit_check_rest.c
```diff
@@ -1239,6 +1244,32 @@ rest_show_check_updates(mtev_http_rest_closure_t *restc,
...
+void
+rest_check_get_attrs(xmlNodePtr attr, char **target, char **name, char **module) {
+  xmlNodePtr a = NULL;
+  for(a = attr->children; a; a = a->next) {
+    if(!strcmp((char *)a->name, "target"))
+      *target = (char *)xmlNodeGetContent(a);
+    if(!strcmp((char *)a->name, "name"))
+      *name = (char *)xmlNodeGetContent(a);
+    if(!strcmp((char *)a->name, "module"))
+      *module = (char *)xmlNodeGetContent(a);
```
bdunavant, line 1256: These should be elseif's, since it's impossible for them to match if the first one did.  

Also, if debug mode is one, should we consider print that there were unknown arttributes as a final else and nothing matched?

src/noit_conf_checks_lmdb.c
```diff
@@ -0,0 +1,205 @@
...
+      else if (data->type == NOIT_LMDB_CHECK_CONFIG_TYPE) {
+        /* We don't care about namespaced config stuff here */
+        if (data->ns == NULL) {
+          char *out_data = (char *)calloc(1, mdb_data.mv_size + 1);
+          memcpy(out_data, mdb_data.mv_data, mdb_data.mv_size);
+          mtev_hash_store(&configh, strdup(data->key), strlen(data->key), out_data);
+        }
+      }
+      else {
+        /* Will hopefully never happen */
```
bdunavant, line 153: should we print warning if it does?

src/noit_check_rest.c
pamaddox: We're going to encountered multiple things that may not match - this is a case where we only want these three.

I added safety to make sure we're setting null for these if nothing matches and added a print for if we couldn't find any of them in the debug log.

src/noit_conf_checks_lmdb.c
pamaddox: Added

src/noit_check_rest.c
pamaddox: Removed the debug prints - "not found" is actually a reasonable/normal outcome for this and we handle those cases.

pamaddox:
Potential thing to look for in review:

I'm not 100% sure I ported all functionality for checks telnet console, since those tended to work directly on the XML config. show and watch both work fine and you can reload all checks from the console. If you see any missing functionality on the console with LMDB that needs addressing, make sure to point it out.

src/noit_lmdb_tools.c
```diff
@@ -0,0 +1,308 @@
...
+
+static mtev_boolean
+lmdb_instance_mkdir(const char *path)
+{
+  char to_make[PATH_MAX];
+  size_t copy_len = strlen(path);
+  memset(to_make, 0, PATH_MAX);
+  memcpy(to_make, path, MIN(copy_len, PATH_MAX));
+  strlcat(to_make, "/dummy", sizeof(to_make));
+  if (mkdir_for_file(to_make, 0777)) {
```
postwait, line 48: 0750, not 0777.

src/noit_lmdb_tools.c
```diff
@@ -0,0 +1,308 @@
...
+  }
+
+  rc = mdb_env_set_maxreaders(env, 1024);
+  if (rc != 0) {
+    errno = rc;
+    mdb_env_close(env);
+    return NULL;
+  }
+
+  rc = mdb_env_open(env, path, 0, 0644);
```
postwait, line 226: 0640 not 0644

src/noit_lmdb_tools.c
```diff
@@ -0,0 +1,308 @@
...
+    mdb_txn_abort(txn);
+    mdb_env_close(env);
+    errno = rc;
+    return NULL;
+  }
+
+  noit_lmdb_instance_t *instance = (noit_lmdb_instance_t *)malloc(sizeof(noit_lmdb_instance_t));
+  instance->env = env;
+  instance->dbi = dbi;
+  ck_rwlock_init(&instance->lock);
```
postwait, line 259: This should be a pthread_wrlock_t...the operations it surrounds are big enough that spinning would be detrimental

src/noit_check_lmdb.c
```diff
@@ -0,0 +1,1571 @@
...
+  mtevAssert(root != NULL);
+  if ((!mdb_data.mv_data) || (mdb_data.mv_size == 0)) {
+    return 0;
+  }
+  if (!strcmp(key_data->key, "uuid")) {
+    /* This should be set separately */
+    return 0;
+  }
+  char *val = (char *)calloc(1, mdb_data.mv_size);
+  memcpy(val, mdb_data.mv_data, mdb_data.mv_size);
```
postwait, line 53: is this null terminated, it doesn't seem that it would be... calloc(1 mdb_data.mv_size+1)? (and future occurences)

src/noit_lmdb_tools.c
pamaddox: Fixed

src/noit_lmdb_tools.c
pamaddox: Fixed

src/noit_lmdb_tools.c
pamaddox: Fixed - uses a `pthread_rwlock_t` now

src/noit_check_lmdb.c
pamaddox: Fixed

PR #600, opened 2020-04-06
CIRC-4638 - Improved Check LMDB On/Off Flags

pamaddox:
Rather than having three flags for having LMDB be on or off:

1: Use LMDB if the "lmdb_path" flag exists and the directory provided
there exists
2: Automatically convert XML checks to LMDB if LMDB is enable


pamaddox:
Comments addressed

PR #607, opened 2020-05-08
CIRC-4638 - Add Ability To Store Filtersets As LMDB

pamaddox:
Add the capability to set an lmdb_path and use_lmdb flag in the filterset
config that will convert the existing XML filtersets to LMDB and will use
LMDB for storage going forward.


pamaddox:
Particular things to look at when reviewing:

* The flatbuffer format/usage
* Making sure we're setting the correct values

src/noit_filters_lmdb.h
```diff
@@ -0,0 +1,58 @@
...
+
+#ifndef _NOIT_FILTERS_LMDB_H
+#define _NOIT_FILTERS_LMDB_H
+
+#include <mtev_rest.h>
+#include <libxml/tree.h>
+
+typedef enum {
+  NOIT_LMDB_FILTERSET_RULE_ACCEPT_TYPE = 'A',
+  NOIT_LMDB_FILTERSET_RULE_ALLOW_TYPE = 'L',
```
postwait, line 38: are these really different? how so?  If they are and both are needed, this would be a great place for a robust comment as to how we ended up needing to differentiate between ACCEPT and ALLOW

pamaddox: This enum wasn't even used - leftover cruft from an older revision of this branch that I eventually scrapped. Removed it.

PR #613, opened 2020-05-14
CIRC-4638 - Check Error Handling And Testing

pamaddox:
Handle checks failing to load more cleanly and add additional testing to
validate that this all works properly.


pamaddox:
Comments are addressed - merging

PR #620, opened 2020-05-18
CIRC-4638 - Add Minimum Time For Filterset Cull

pamaddox:
Do not cull a filterset unless it has been idle for a configurable
ammount of time (default: 5 minutes).


pamaddox:
That's more difficult than it looks. The reason I did it this way is - the way we determine that a filterset is ready to be culled is if it exists in the configuration/DB, but not in memory. By definition, if that's the case, the "last_used_s" field wouldn't exist, since it would already be removed from the DB. This was the path of least resistance to getting this working.

If we want to switch to add a field to the filterset_t, we'll need to change how the filterset cull mechanism works; it's doable, but that may take a while longer.

postwait:
We should change it to only cull those that are in memory -- that seems like it would solve this issue.

pamaddox:
Performance of this will be too slow.... closing this, will need an alternate approach.

src/noit_filters.c
```diff
@@ -101,10 +106,60 @@ noit_filter_initialized() {
   return initialized;
 }
 void
+noit_filter_update_last_touched(const char *filterset_name) {
+  struct timeval *now = (struct timeval *)calloc(1, sizeof(struct timeval));
+  mtev_gettimeofday(now, NULL);
+  LOCKFSLT();
+  mtev_hash_replace(filterset_last_touched, strdup(filterset_name), strlen(filterset_name), (void *)now, free, free);
```
pjulien, line 113: I am new to these implementations but can we not allocate and free this memory outside of the the critical section at least?

There's no need to duplicate the string while the lock is held, i don't know if the call to ``free`` can be done outside the lock since it seems to be embedded in the call.

postwait: No reason.  This whole PR was scraped due to performance concerns.

PR #622, opened 2020-05-20
Support a limit for queue backlogs.

postwait:


HeinrichHartmann:
LGTM so far. Only question is how we set `drop_backlog_over` from caql-broker. Probably best to expose this to lua, no?

HeinrichHartmann:
perfect. LGTM.

PR #623, opened 2020-05-20
CIRC-4638 - Add filters/show Endpoint

pamaddox:
Add a new endpoint - "filters/show", that functions like the existing "checks/show"
endpoint. It returns an XML list of all filtersets.


src/noit_filters_lmdb.c
```diff
@@ -1231,6 +1231,69 @@ noit_filters_lmdb_process_repl(xmlDocPtr doc) {
...
+  goto cleanup;
+
+ error:
+  mtev_http_response_standard(ctx, error_code, "ERROR", "text/html");
+  mtev_http_response_end(ctx);
+  goto cleanup;
+
+ cleanup:
+  if(doc) {
+    xmlFreeDoc(doc);
```
pjulien, line 1291: Is this reference counted or does ``mtev_http_response_xml`` render the entire document to a string?

pamaddox: `mtev_http_response_xml` returns a string to the caller

src/noit_filters_lmdb.c
```diff
@@ -1231,6 +1231,69 @@ noit_filters_lmdb_process_repl(xmlDocPtr doc) {
...
+
+  mtev_http_response_ok(ctx, "text/xml");
+  mtev_http_response_xml(ctx, doc);
+  mtev_http_response_end(ctx);
+  goto cleanup;
+
+ error:
+  mtev_http_response_standard(ctx, error_code, "ERROR", "text/html");
+  mtev_http_response_end(ctx);
+  goto cleanup;
```
vasuraman11, line 1287: nit - unneeded, cleanup is the next source line

pamaddox: Here for consistency. The existing endpoints also do this.

PR #629, opened 2020-05-29
WIP: Metric Director Move Account Subscriptions Logic

HeinrichHartmann:
This is some prep work, for improving the tag-search routing logic.

The account subscription logic was mis-placed in the noit_lua_binding_* files.
Moving those to the metric_director makes more sense, and also avoids quirks
with the initialization logic.

However, this project is not complete.

- Caql-broker tests are flapping with this. There seem to be issues with the delivery of the messages. But it's not clear to me what's causing this. The code change should be feature neutral.

- Debugging is complicated by the fact that mtevL() does not work from the hook handler. #BUG

### DIFF
--- a/src/noit_metric_director.c
+++ b/src/noit_metric_director.c
@@ -83,6 +83,8 @@ static volatile void **thread_queues;
 static uint32_t *thread_queues_backlog;
 static mtev_hash_table id_level;
 static mtev_hash_table dedupe_hashes;
+static mtev_hash_table account_interests_map;
+
 static caql_cnt_t *check_interests;
 static mtev_boolean dedupe = mtev_true;
 static uint32_t director_in_use = 0;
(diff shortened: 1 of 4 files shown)


src/noit_metric_director.c
```diff
@@ -137,15 +139,52 @@ noit_metric_director_my_lane() {
...
+      break;
+    }
+  }
+  mtevAssert(account_interests);
+  thread_id = get_my_lane();
+  icnt = account_interests[thread_id];
+  icnt += cnt;
+  if (icnt < 0) icnt = 0;
+  mtevAssert(icnt <= 0xffff);
+  account_interests[thread_id] = icnt;
```
pjulien, line 177: I read this file last Friday prior to this PR and I've seen this pattern in a few places already... and I don't understand how this is thread safe.  Nor can I can see how it possibly can be.

Can someone enlighten on this please?  What is keeping ``account_interests`` safe?  Because I do not see it.

postwait: I believe it is not thread safe and it assumes that you can only ever operate on your thread.  If this assumption is not true, it is broken.

pjulien: @postwait 

It's a static variable in the presence of multiple threads.  Then it's not safe to use.  The assumption seems to be that it's safe because of the index accessed is based on ``thread_id``.

This is incorrect because of false sharing.  For this to work properly, we need at a minimum one cache line of space between the "thread slots"

PR #635, opened 2020-06-08
 CIRC-4638 - Fix Empty XML LMDB Conversion

pamaddox:
We were skipping empty XML elements when converting from XML to LMDB.
Make sure we store these as keys will null values in the DB rather than
skipping them. Add tests to confirm that this works.

### DIFF
--- a/src/noit_check_lmdb.c
+++ b/src/noit_check_lmdb.c
@@ -50,7 +50,10 @@ static int noit_check_lmdb_add_attribute(xmlNodePtr root, xmlNodePtr attr, noit_
   if ((mdb_data.mv_data == NULL) && (mdb_data.mv_size > 0)) {
     /* If data is null, but we have a size, something is wrong - just skip
      * this key */
-    mtevL(mtev_error, "noit_check_lmdb_add_attribute: got null data for key with a size, skipping\n");
+    char uuid_str[UUID_STR_LEN + 1];
+    mtev_uuid_unparse_lower(key_data->id, uuid_str);
+    mtevL(mtev_error, "noit_check_lmdb_add_attribute: got null data for key with a size - check %s, namespace %s, key %s - skipping\n",
+        uuid_str, (key_data->ns) ? key_data->ns : "<null>", key_data->key);
     return 0;
   }
   if ((mdb_data.mv_data != NULL) && (mdb_data.mv_size > 0)) {
@@ -77,8 +80,10 @@ static int noit_check_lmdb_add_config(xmlNodePtr root, xmlNodePtr config, noit_l
   if ((mdb_data.mv_data == NULL) && (mdb_data.mv_size > 0)) {
     /* If data is null, but we have a size, something is wrong - just skip
      * this key */
-    mtevL(mtev_error, "noit_check_lmdb_add_config: got null data for key with a size, skipping\n");
-    return 0;
+    char uuid_str[UUID_STR_LEN + 1];
+    mtev_uuid_unparse_lower(key_data->id, uuid_str);
+    mtevL(mtev_error, "noit_check_lmdb_add_config: got null data for key with a size - check %s, namespace %s, key %s - skipping\n",
+        uuid_str, (key_data->ns) ? key_data->ns : "<null>", key_data->key);
   }
   if ((mdb_data.mv_data != NULL) && (mdb_data.mv_size > 0)) {
     val = (char *)calloc(1, mdb_data.mv_size + 1);
--- a/src/noit_check_lmdb.c
+++ b/src/noit_check_lmdb.c
@@ -42,15 +42,21 @@ static int noit_check_lmdb_show_check_json(mtev_http_rest_closure_t *restc,
 
 static int noit_check_lmdb_add_attribute(xmlNodePtr root, xmlNodePtr attr, noit_lmdb_check_data_t *key_data, MDB_val mdb_data, bool separate_stanza) {
   mtevAssert(root != NULL);
-  if ((!mdb_data.mv_data) || (mdb_data.mv_size == 0)) {
-    return 0;
-  }
+  char *val = NULL;
   if (!strcmp(key_data->key, "uuid")) {
     /* This should be set separately */
     return 0;
   }
-  char *val = (char *)calloc(1, mdb_data.mv_size + 1);
-  memcpy(val, mdb_data.mv_data, mdb_data.mv_size);
+  if ((mdb_data.mv_data == NULL) && (mdb_data.mv_size > 0)) {
+    /* If data is null, but we have a size, something is wrong - just skip
+     * this key */
+    mtevL(mtev_error, "noit_check_lmdb_add_attribute: got null data for key with a size, skipping\n");
+    return 0;
+  }
+  if ((mdb_data.mv_data != NULL) && (mdb_data.mv_size > 0)) {
+    val = (char *)calloc(1, mdb_data.mv_size + 1);
+    memcpy(val, mdb_data.mv_data, mdb_data.mv_size);
+  }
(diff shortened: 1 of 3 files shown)


src/noit_check_lmdb.c
```diff
@@ -42,15 +42,21 @@ static int noit_check_lmdb_show_check_json(mtev_http_rest_closure_t *restc,
...
   if (!strcmp(key_data->key, "uuid")) {
     /* This should be set separately */
     return 0;
   }
-  char *val = (char *)calloc(1, mdb_data.mv_size + 1);
-  memcpy(val, mdb_data.mv_data, mdb_data.mv_size);
+  if ((mdb_data.mv_data == NULL) && (mdb_data.mv_size > 0)) {
+    /* If data is null, but we have a size, something is wrong - just skip
+     * this key */
+    mtevL(mtev_error, "noit_check_lmdb_add_attribute: got null data for key with a size, skipping\n");
```
esproul, line 53: If there is anything useful we can log here, like the name of the attribute, or anything else that can make this error actionable by an operator/user, that would be helpful.

src/noit_check_lmdb.c
```diff
@@ -67,11 +73,17 @@ static int noit_check_lmdb_add_attribute(xmlNodePtr root, xmlNodePtr attr, noit_
 static int noit_check_lmdb_add_config(xmlNodePtr root, xmlNodePtr config, noit_lmdb_check_data_t *key_data, MDB_val mdb_data) {
   mtevAssert(root != NULL);
   mtevAssert(config != NULL);
-  if ((!mdb_data.mv_data) || (mdb_data. mv_size == 0)) {
+  char *val = NULL;
+  if ((mdb_data.mv_data == NULL) && (mdb_data.mv_size > 0)) {
+    /* If data is null, but we have a size, something is wrong - just skip
+     * this key */
+    mtevL(mtev_error, "noit_check_lmdb_add_config: got null data for key with a size, skipping\n");
```
esproul, line 80: Same here, maybe tell us what key it was, or whatever could be used to track this down.

src/noit_check_lmdb.c
```diff
@@ -42,15 +42,21 @@ static int noit_check_lmdb_show_check_json(mtev_http_rest_closure_t *restc,
...
-  memcpy(val, mdb_data.mv_data, mdb_data.mv_size);
+  if ((mdb_data.mv_data == NULL) && (mdb_data.mv_size > 0)) {
+    /* If data is null, but we have a size, something is wrong - just skip
+     * this key */
+    mtevL(mtev_error, "noit_check_lmdb_add_attribute: got null data for key with a size, skipping\n");
+    return 0;
+  }
+  if ((mdb_data.mv_data != NULL) && (mdb_data.mv_size > 0)) {
+    val = (char *)calloc(1, mdb_data.mv_size + 1);
+    memcpy(val, mdb_data.mv_data, mdb_data.mv_size);
```
vasuraman11, line 58:
```suggestion
    memcpy(val, mdb_data.mv_data, mdb_data.mv_size);
    val[mdb_data.mv_size] = 0;
```

src/noit_check_lmdb.c
```diff
@@ -67,11 +73,17 @@ static int noit_check_lmdb_add_attribute(xmlNodePtr root, xmlNodePtr attr, noit_
...
+    /* If data is null, but we have a size, something is wrong - just skip
+     * this key */
+    mtevL(mtev_error, "noit_check_lmdb_add_config: got null data for key with a size, skipping\n");
     return 0;
   }
-  char *val = (char *)calloc(1, mdb_data.mv_size + 1);
-  memcpy(val, mdb_data.mv_data, mdb_data.mv_size);
+  if ((mdb_data.mv_data != NULL) && (mdb_data.mv_size > 0)) {
+    val = (char *)calloc(1, mdb_data.mv_size + 1);
+    memcpy(val, mdb_data.mv_data, mdb_data.mv_size);
```
vasuraman11, line 85:
```suggestion
    memcpy(val, mdb_data.mv_data, mdb_data.mv_size);
    val[mdb_data.mv_size] = 0;
```

src/noit_check_lmdb.c
pamaddox: We don't need to do that. We're using `calloc`, which automatically memsets the data to zero.

src/noit_check_lmdb.c
pamaddox: We don't need to do that. We're using calloc, which automatically memsets the data to zero.

src/noit_check_lmdb.c
```diff
@@ -67,11 +73,17 @@ static int noit_check_lmdb_add_attribute(xmlNodePtr root, xmlNodePtr attr, noit_
...
+  char *val = NULL;
+  if ((mdb_data.mv_data == NULL) && (mdb_data.mv_size > 0)) {
+    /* If data is null, but we have a size, something is wrong - just skip
+     * this key */
+    mtevL(mtev_error, "noit_check_lmdb_add_config: got null data for key with a size, skipping\n");
     return 0;
   }
-  char *val = (char *)calloc(1, mdb_data.mv_size + 1);
-  memcpy(val, mdb_data.mv_data, mdb_data.mv_size);
+  if ((mdb_data.mv_data != NULL) && (mdb_data.mv_size > 0)) {
```
vasuraman11, line 83: no else here what happens if it is null and zero or if it is non-null but zero size but something is there - can that happen?  Note that above you had one case when it was NULL where you malloc'ed a single byte into value and and put a zero into it...

test/busted/simple/check_config/check_config_spec.lua
```diff
@@ -0,0 +1,153 @@
...
+  local noit, api
+  setup(function()
+    Reconnoiter.clean_workspace()
+    noit = Reconnoiter.TestNoit:new()
+  end)
+  teardown(function() if noit ~= nil then noit:stop() end end)
+
+  local uuid = mtev.uuid()
+  local check_xml = function(seq)
+    return
```
vasuraman11, line 11: this returns here but then has a string right afterwards, maybe need a comment here to explain

src/noit_check_lmdb.c
pamaddox: Fixed

src/noit_check_lmdb.c
pamaddox: Fixed

src/noit_check_lmdb.c
vasuraman11: ah ok guess maybe it was malloc in the other case where you explicitly set last byte to 0

test/busted/simple/check_config/check_config_spec.lua
pamaddox: This exact pattern is used extremely frequently throughout the test suite. It's doing exactly what you think - it's returning the string immediately following the "return". If we want to comment this, we should do it systematically throughout the test suite rather than just here.

src/noit_check_lmdb.c
vasuraman11: kk

src/noit_check_lmdb.c
pamaddox: There's no else because val is already set to NULL, which is what we want. This case is covered in the test suite.

test/busted/simple/check_config/check_config_spec.lua
vasuraman11: ah ok, wasn't used to seeing the return value on the next line

src/noit_check_lmdb.c
vasuraman11: ok

PR #637, opened 2020-06-12
Facilities to register searches to director lanes

postwait:
* Make the direct interests thread safe (cross-thread registration)
* Support registering searches with a lane
* Simplify the distribution path
* Make noit_metric_search_tag_ast_t ref counted.

### DIFF
--- a/src/noit_metric_director.c
+++ b/src/noit_metric_director.c
@@ -88,13 +88,19 @@ typedef union {
 } thread_queue_t;
 
 static int nthreads;
+
+static inline int safe_thread_id(int thread_id) {
+  if(thread_id < 0) thread_id ^= thread_id;
+  thread_id = thread_id % nthreads;
+  return thread_id;
+}
+
 static thread_queue_t *queues;
 static struct {
   ck_spinlock_t lock;
   ck_hs_t hs;
 } *search_asts;
 static mtev_hash_table id_level;
-static pthread_mutex_t id_level_lock;
 static mtev_hash_table dedupe_hashes;
 static interest_cnt_t *check_interests;
 static pthread_mutex_t check_interests_lock;
@@ -166,8 +172,7 @@ noit_metric_director_adjust_checks_interest_on_thread(int thread_id, short cnt) 
   interest_cnt_t *src, *dst;
   int icnt, bytes;
 
-  if(thread_id < 0) thread_id ^= thread_id;
-  thread_id = thread_id % nthreads;
+  thread_id = safe_thread_id(thread_id);
 
   mtev_memory_begin();
   pthread_mutex_lock(&check_interests_lock);
@@ -189,44 +194,43 @@ noit_metric_director_adjust_checks_interest_on_thread(int thread_id, short cnt) 
 
 /*
  * the ASTs... each lane can register a set of ASTs
- * We are doing the concurrently, so we make a copy of the ASTs.
+ * We are doing this concurrently, so we make a copy of the ASTs.
  *
  * [ thread_id, ck_hs [ ID: AST ] ]
  *
  */
 
 static uint32_t ast_counter = 1;
-struct search2 {
+typedef struct {
   uint32_t    ast_id; /* unique identifier for removal */
   int64_t account_id; /* >= 0 -> exact match, < 0 -> allow all */
   uuid_t  check_uuid; /* UUID_ZERO -> any, otherwise exact match */
   noit_metric_tag_search_ast_t *ast;
-};
+} search2_t;
 
 static void
 search2_cleanup(void *v) {
-  struct search2 *s = (struct search2 *)v;
+  search2_t *s = (search2_t *)v;
   noit_metric_tag_search_free(s->ast);
 }
 
 static unsigned long
 search2_hash(const void *object, unsigned long seed) {
-  struct search2 *s = (struct search2 *)object;
+  search2_t *s = (search2_t *)object;
   return s->ast_id * seed;
 }
 static bool
 search2_compare(const void *previous, const void *compare) {
-  return ((struct search2 *)previous)->ast_id == ((struct search2 *)compare)->ast_id;
+  return ((search2_t *)previous)->ast_id == ((search2_t *)compare)->ast_id;
 }
 
 uint32_t
 noit_metric_director_register_search_on_thread(int thread_id, int64_t account_id, uuid_t check_uuid,
                                noit_metric_tag_search_ast_t *ast) {
   /* Here we want to add the AST to a specific thread's interest. */
-  if(thread_id < 0) thread_id ^= thread_id;
-  thread_id = thread_id % nthreads;
+  thread_id = safe_thread_id(thread_id);
   mtev_memory_begin();
-  struct search2 *as = mtev_memory_safe_malloc_cleanup(sizeof(*as), search2_cleanup);
+  search2_t *as = mtev_memory_safe_malloc_cleanup(sizeof(*as), search2_cleanup);
   memset(as, 0, sizeof(*as));
   as->ast_id = ck_pr_faa_32(&ast_counter, 1);
   as->account_id = account_id;
@@ -247,10 +251,9 @@ noit_metric_director_register_search(int64_t account_id, uuid_t check_uuid, noit
 
 mtev_boolean
 noit_metric_director_deregister_search_on_thread(int thread_id, uint32_t ast_id) {
-  if(thread_id < 0) thread_id ^= thread_id;
-  thread_id = thread_id % nthreads;
+  thread_id = safe_thread_id(thread_id);
   mtev_memory_begin();
-  struct search2 *found, dummy = { .ast_id = ast_id };
+  search2_t *found, dummy = { .ast_id = ast_id };
   unsigned long hash = CK_HS_HASH(&search_asts[thread_id].hs, search2_hash, &dummy);
   ck_spinlock_lock(&search_asts[thread_id].lock);
   found = ck_hs_remove(&search_asts[thread_id].hs, hash, &dummy);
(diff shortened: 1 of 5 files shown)


src/noit_metric_director.c
```diff
@@ -280,47 +413,70 @@ noit_metric_director_flush(eventer_t e) {
...
+  /* First we process interests for specific metrics: uuid-metric|ST[tags...] */
+  for(i=0; i<2; i++) {
+    /* First time through we use the message's ID, second time we use the NULL UUID
+     * which is used to "match any" check.
+     */
+    if(mtev_hash_retrieve(&id_level, i ? (const char *)&uuid_zero : (const char *)&message->id.id, UUID_SIZE,
+        &vhash)) {
+      if(mtev_hash_retrieve((mtev_hash_table *) vhash, message->id.name,
+                            message->id.name_len_with_tags, (void **)&vinterests)) {
+        memcpy(interests, *vinterests, sizeof(interests)); /*  we access this exactly once, no need to ck load or tmp variable */
```
pjulien, line 434: For this line and 439, the code is correct... but the comment is not.  What makes it safe is 

1. SMR
1. That we'll see either the old pointer of the array or the new one.

In other words, we're not using a fence to enforce visibility but we're OK with just atomicity.

src/noit_metric_director.c
```diff
@@ -280,47 +413,70 @@ noit_metric_director_flush(eventer_t e) {
...
+        for(i=0; i<nthreads; i++) has_interests |= interests[i];
+      }
+      if(mtev_hash_retrieve((mtev_hash_table *) vhash, message->id.name,
+                            message->id.name_len, (void **)&vinterests)) {
+        memcpy(interests, *vinterests, sizeof(interests)); /*  we access this exactly once, no need to ck load or tmp variable */
+        for(i=0; i<nthreads; i++) has_interests |= interests[i];
```
pjulien, line 440: What's the visibility rule in C again for this?  Is this the same `i` than on line 426/435?

I believe the entire body of the for loop should be its own function anyway considering the importance of this code.

src/noit_metric_director.c
```diff
@@ -280,47 +413,70 @@ noit_metric_director_flush(eventer_t e) {
...
+      }
     }
   }
 
+  /* Next we run search queries per lane to find matches */
+  for(i=0; i<nthreads; i++) {
+    if(interests[i] != 0) continue; /* already interest, skip any work */
+    struct search2 *search;
+    ck_hs_iterator_t iter = CK_HS_ITERATOR_INITIALIZER;
+    while(ck_hs_next_spmc(&search_asts[i].hs, &iter, (void **)&search)) {
```
pjulien, line 450: wasn't a spin lock put in place to protect this hash table?

src/noit_metric_director.c
```diff
@@ -640,11 +796,19 @@ void noit_metric_director_init() {
...
+    ck_hs_init(&search_asts[i].hs, CK_HS_MODE_OBJECT | CK_HS_MODE_SPMC,
+               search2_hash, search2_compare,
+               &mtev_memory_safe_ck_malloc, 128, mtev_rand());
+  }
 
   mtev_memory_begin();
   check_interests = mtev_memory_safe_calloc(sizeof(*check_interests),nthreads);
   mtev_memory_end();
   pthread_mutex_init(&check_interests_lock, NULL);
+  pthread_mutex_init(&id_level_lock, NULL);
```
pjulien, line 811: `id_level_lock` is no longer used anywhere except here

src/noit_metric_director.h
```diff
@@ -35,12 +35,14 @@
...
 #include <mtev_uuid.h>
 #include <noit_message_decoder.h>
+#include <noit_metric_tag_search.h>
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-typedef unsigned short caql_cnt_t;
+typedef unsigned short interest_cnt_t;
```
pjulien, line 44: if this is unsigned why are there tests in the implementation file checking if it's under 0?  The value would need to be cast or checked using bits to see if it's in the negative range

src/noit_metric_director.h
```diff
@@ -64,12 +66,20 @@ void noit_metric_director_dedupe(mtev_boolean dedupe);
 
 /* Tells noit to funnel all observed lines matching this id-metric
  * back to this thread */
-caql_cnt_t noit_adjust_metric_interest(uuid_t id, const char *metric, short cnt);
+interest_cnt_t noit_adjust_metric_interest(uuid_t id, const char *metric, short cnt) __attribute__((deprecated));
+interest_cnt_t noit_metric_director_adjust_metric_interest_on_thread(int thread_id, uuid_t id, const char *metric, short cnt);
+interest_cnt_t noit_metric_director_adjust_metric_interest(uuid_t id, const char *metric, short cnt);
 
+uint32_t noit_metric_director_register_search_on_thread(int thread_id, int64_t account_id, uuid_t check_uuid, noit_metric_tag_search_ast_t *ast);
```
pjulien, line 73: you sure you want to make this call public?

src/noit_metric_director.h
```diff
@@ -64,12 +66,20 @@ void noit_metric_director_dedupe(mtev_boolean dedupe);
...
+uint32_t noit_metric_director_register_search(int64_t account_id, uuid_t check_uuid, noit_metric_tag_search_ast_t *ast);
+mtev_boolean noit_metric_director_deregister_search_on_thread(int thread_id, uint32_t ast_id);
```
pjulien, line 75: you sure you want to make this call public?

src/noit_metric_tag_search.c
```diff
@@ -37,10 +37,12 @@
 #include "noit_metric_tag_search.h"
 
 #include <stdio.h>
+#include <ck_pr.h>
 
 void
 noit_metric_tag_search_free(noit_metric_tag_search_ast_t *node) {
   if(node == NULL) return;
+  if(!ck_pr_dec_32_is_zero(&node->refcnt)) return;
```
pjulien, line 45: if it's not freed here, couldn't a zero here prevent it from eventually being freed?

postwait: If there is a zero here, it would mean it has been freed and this would be a double free.

src/noit_metric_director.h
postwait: Yes, required for fault.

src/noit_metric_director.h
postwait: Yes, required for fault.

src/noit_metric_director.h
postwait: We are applying checks to a larger type to ensure it doesn't roll.

src/noit_metric_director.c
postwait: removed

src/noit_metric_director.c
```diff
@@ -170,14 +187,101 @@ noit_adjust_checks_interest(short cnt) {
...
+}
+
+uint32_t
+noit_metric_director_register_search_on_thread(int thread_id, int64_t account_id, uuid_t check_uuid,
+                               noit_metric_tag_search_ast_t *ast) {
+  /* Here we want to add the AST to a specific thread's interest. */
+  if(thread_id < 0) thread_id ^= thread_id;
+  thread_id = thread_id % nthreads;
+  mtev_memory_begin();
+  struct search2 *as = mtev_memory_safe_malloc_cleanup(sizeof(*as), search2_cleanup);
```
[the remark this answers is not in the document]
postwait: the mtev_memory_safe_free function will free it.  The cleanup is any additional cleanup that is required.

src/noit_metric_director.c
```diff
@@ -170,14 +187,101 @@ noit_adjust_checks_interest(short cnt) {
...
+};
+
+static void
+search2_cleanup(void *v) {
+  struct search2 *s = (struct search2 *)v;
+  noit_metric_tag_search_free(s->ast);
+}
+
+static unsigned long
+search2_hash(const void *object, unsigned long seed) {
```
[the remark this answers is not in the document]
postwait: This is a function prototype match for ck_hs.

src/noit_metric_director.c
```diff
@@ -170,14 +187,101 @@ noit_adjust_checks_interest(short cnt) {
...
+  if(thread_id < 0) thread_id ^= thread_id;
+  thread_id = thread_id % nthreads;
+  mtev_memory_begin();
+  struct search2 *as = mtev_memory_safe_malloc_cleanup(sizeof(*as), search2_cleanup);
+  memset(as, 0, sizeof(*as));
+  as->ast_id = ck_pr_faa_32(&ast_counter, 1);
+  as->account_id = account_id;
+  mtev_uuid_copy(as->check_uuid, check_uuid);
+  as->ast = noit_metric_tag_search_ref(ast);
+  unsigned long hash = CK_HS_HASH(&search_asts[thread_id].hs, search2_hash, as);
```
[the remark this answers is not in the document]
postwait: agreed. Will fix.

src/noit_metric_director.c
```diff
@@ -187,38 +291,67 @@ noit_adjust_metric_interest(uuid_t id, const char *metric, short cnt) {
     mtev_hash_init_locks(level2, MTEV_HASH_DEFAULT_SIZE, MTEV_HASH_LOCK_MODE_MUTEX);
     mtev_uuid_copy(*copy, id);
     if(!mtev_hash_store(&id_level, (const char *)copy, UUID_SIZE, level2)) {
-      free(copy);
+      mtev_hash_destroy(level2, NULL, NULL);
```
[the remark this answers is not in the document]
postwait: yes

src/noit_metric_director.c
```diff
@@ -187,38 +291,67 @@ noit_adjust_metric_interest(uuid_t id, const char *metric, short cnt) {
...
+  interest_cnt_t *newinterests = NULL;
+  do {
+    if(newinterests) mtev_memory_safe_free(newinterests);
+    interests = ck_pr_load_ptr(vinterests);
+    /* This is fine because thread_id is only ours */
+    icnt = interests[thread_id];
+    icnt += cnt;
+    if(icnt < 0) icnt = 0;
+    mtevAssert(icnt <= 0xffff);
+    interest_cnt_t *newinterests = mtev_memory_safe_calloc(nthreads, sizeof(*newinterests));
```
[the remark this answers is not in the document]
postwait: We'd need locks on the read side if we didn't use a new region leaving the old in-tact.

src/noit_metric_director.c
postwait: yes... for modification

src/noit_metric_director.c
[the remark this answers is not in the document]
pjulien: No, you don't, you just calloc before the start of the loop

PR #643, opened 2020-06-24
Allow search interests so specify exact metric_name

postwait:
* This avoids work for search patterns that have exact metric name
  matching.
* This also simplifies the logic a bit by not having thread_asts at
  multiple levels of the heirarchy by explicitly storing the "any check"
  as the null uuid and "any metric name" as the "" blank name thus
  leveraging the normal check_search_t and name_search_t retrieval
  mechanism.

### DIFF
--- a/src/noit_metric_director.c
+++ b/src/noit_metric_director.c
@@ -98,16 +98,19 @@
  *
  *    Each account has an `account_search_t` structure that has a generational counter
  *    that is modified any time a search expression is added. The structure has a
- *    hash table of registered searches for each lane that apply to "all checks"
- *    (NULL UUID) and a hash table of check_search_t structures indexed by UUID that
- *    contain a hash table per lane of all registered searches that apply to exactly
- *    one check.
+ *    hash table of check_search_t structures indexed by UUID that (NULL UUID meaning
+ *    and check matchs) contain a hash of name_search_t by name (where NULL/"" means
+ *    it applies to any name) which each contain a table per lane of all registered
+ *    searches that apply.
  *
  *    To evaluate a given inbound metric, we first check the miss cache and if the
  *    item is present in the miss cache, then we skip all work.  Otherwise, 
- *    the account_search_t is found for the appropriate account and the "all checks"
- *    searches are applied.  Then the check specific set of searches are applied. As
- *    soon as a lane has a single match all subsequent searches are skipped.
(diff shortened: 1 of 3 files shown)


src/noit_metric_director.c
```diff
@@ -98,16 +98,19 @@
...
  *    Each account has an `account_search_t` structure that has a generational counter
  *    that is modified any time a search expression is added. The structure has a
- *    hash table of registered searches for each lane that apply to "all checks"
- *    (NULL UUID) and a hash table of check_search_t structures indexed by UUID that
- *    contain a hash table per lane of all registered searches that apply to exactly
- *    one check.
+ *    hash table of check_search_t structures indexed by UUID that (NULL UUID meaning
+ *    and check matchs) contain a hash of name_search_t by name (where NULL/"" means
+ *    it applies to any name) which each contain a table per lane of all registered
+ *    searches that apply.
```
esproul, line 104:
```suggestion
 *    hash table of check_search_t structures indexed by UUID (where the NULL UUID means
 *    any check matches) that contain a hash of name_search_t by name (where NULL/"" 
 *    means it applies to any name), each of which contains a table per lane of all registered
 *    searches that apply.
```

src/noit_metric_director.c
```diff
@@ -461,48 +515,42 @@ noit_metric_director_register_search_on_thread(int thread_id, int64_t account_id
...
-    ck_spinlock_unlock(&searches->search_asts[thread_id].lock);
-  } else {
-    check_search_t *cs = check_get_search(searches, check_uuid, true);
-    unsigned long hash = CK_HS_HASH(&cs->search_asts[thread_id].hs, tag_search_registration_hash, as);
-    ck_spinlock_lock(&cs->search_asts[thread_id].lock);
-    ck_hs_put(&cs->search_asts[thread_id].hs, hash, as);
-    ck_spinlock_unlock(&cs->search_asts[thread_id].lock);
-  }
+
+  check_search_t *cs = check_get_search(searches, check_uuid, true);
```
pjulien, line 519: What is the reason that `check_uuid` cannot be null anymore?

postwait: The null UUID is a misnomer -- it is actually a UUID with all zeros.  And now it is explcitly stored in the checks hash as such.

It is ever-so-slightly less efficient, but makes the structure much easier to understand.

PR #645, opened 2020-07-23
Make the version cleaner and avoid snprintf.

postwait:
* Expose three global symbols to aide debuggers.

### DIFF
--- a/buildtools/mkversion.sh
+++ b/buildtools/mkversion.sh
@@ -9,12 +9,12 @@ if [ $? -eq 0 ]; then
   SYM=`git symbolic-ref -q HEAD | awk -F'/' '{ print $NF }'`
   if [ -z "$SYM" ]; then
     SYM="detached"
-  fi
-  if [ -z "`echo $SYM | grep '^tags/'`" ]; then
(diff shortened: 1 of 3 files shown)


buildtools/mkversion.sh
```diff
@@ -9,12 +9,12 @@ if [ $? -eq 0 ]; then
...
   if [ -z "$SYM" ]; then
     SYM="detached"
-  fi
-  if [ -z "`echo $SYM | grep '^tags/'`" ]; then
+  elif [ -z "`echo $SYM | grep '^tags/'`" ]; then
     SYM="branches/$SYM"
   fi
   echo "    * symbolic -> $SYM"
   BRANCH=$SYM
+  VSTR=`echo -n "$BRANCH" | sed -e 's#^tags/##;' | sed -e 's#^branches/##;'`
```
esproul, line 17: Same issue as in libmtev with echo. Should use printf here.
```suggestion
  VSTR=`printf "$BRANCH" | sed -e 's#^tags/##;' | sed -e 's#^branches/##;'`
```

PR #646, opened 2020-07-23
Improve time encoding on B* records and IEP filtering

postwait:
* Make IEP filtering apply to B* and H* records.
* Change the default filter to 5 minutes and make configurable.
* When encoding a set of metrics, if the timestamps are all set then
  rewind the encapulating line's timestamp to the latest of those.
* Alter the BF format before it is put into use to include a timestamp.

### DIFF
--- a/src/noit_check_log.c
+++ b/src/noit_check_log.c
@@ -318,7 +318,7 @@ noit_check_log_bundle_metric_flatbuffer_serialize_log(mtev_log_stream_t ls,
 static int
 noit_check_log_bundle_metric_serialize(mtev_log_stream_t ls,
                                        noit_check_t *check,
-                                       const struct timeval *whence,
+                                       const struct timeval *in_whence,
                                        metric_t *m) {
   int size, rv = -1;
   unsigned int out_size;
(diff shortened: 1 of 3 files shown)


src/noit_check_log.c
```diff
@@ -628,14 +638,22 @@ noit_check_log_bundle_fb_serialize(mtev_log_stream_t ls, noit_check_t *check, st
         rv_err = -1;
       } else {
         int rv;
+        struct timeval *time_to_use = whence;
+        /* If whence is unset OR if laste_metric is set and less than whence,
```
pjulien, line 642: typo most likely

src/noit_metric_director.c
```diff
@@ -1027,8 +1027,14 @@ handle_metric_buffer(const char *payload, int payload_len,
         stats_add64(stats_msg_seen, 1);
         int rv = noit_message_decoder_parse_line(message, has_noit);
 
-        if(message->value.whence_ms) // ignore screwy messages
-          stats_set_hist_intscale(stats_msg_delay, mtev_now_ms() - message->value.whence_ms, -3, 1);
+        if(message->value.whence_ms) {
+          // ignore screwy messages
+          int64_t diff = mtev_now_ms() - message->value.whence_ms;
+          if(diff > 300000) {
+            mtevL(mtev_error, "[%zd] %.*s\n", diff, payload_len, copy);
```
esproul, line 1034: This doesn't look like a particularly actionable errorlog message. Is it more of a debug thing perhaps?

postwait: this was debugging, not supposed to be part of this.

PR #652, opened 2020-10-22
Call module cleanup immediately on removal.

postwait:
This requires module cleanup functions to be safe when run multiple
times.  This fixes lingering sockets on checks that have listeners
which is important when the check is modified but the port is unchanged
because the stale check will still have an active listener, but will
not handle arriving data which can cause loss.


bdunavant (approved):
LGTM

PR #656, opened 2020-11-03
Fix bare tags and statsd DD tag decoding

postwait:
* Allow decoding of bare tag names `foo` as `foo:`
* Fix decoding of DD-style statsd tags in multi-line mode.


bdunavant (approved):
lgtm

pjulien (approved):

PR #662, opened 2020-12-03
Protect hisoteir with locks and make hashes memory safe

postwait:


vasuraman11 (approved):
LGTM cool, locking around that

PR #664, opened 2020-12-07
reopen log files with dropped privileges.

postwait:
In noitd child_main before dropping privileges we call log reopen which,
conceptually, would not alter/create anything.  However, if there is a
jlog precommit buffer and the last segment is full, it will cause a
flush which can cause a newly created segment owned by the wrong user.

Call instead mtev_conf_log_init which will reinitializing the logging
system under dropped effective privileges.


esproul:
Are the `config_file` changes to prevent NPE?

postwait:
The `config_file` changes prevent reassessment of the path post chdir

esproul (approved):
LGTM

PR #665, opened 2020-12-10
SNMP tagging

rileyberton:
Adds support for sending stream tags associated with an OID in
the SNMP config

```
<config>
  <oid_foo>.1.2.3</oid_foo>
  <tags_foo>baz:bing</tags_foo>
  <type_foo>uint32</type_foo>
</config>
```

Without the `<tags_.+>` config field, it uses plain 'foo', otherwise:

`foo|ST[baz:bing]`


rileyberton:
This won't do what I expect, closing this.

PR #671, opened 2021-01-07
Allow foo|ST[] as a valid metric name.

postwait:
Fix an off-by-one-error that refused empty tag sets in canonicalization.


bdunavant (approved):
LGTM

PR #675, opened 2021-01-12
CIRC-6121 - Process Statsd Results In Parallel

pamaddox:
Allow configuring statsd to process entries in parallel rather than
doing them all in a single thread.

### DIFF
--- a/src/modules/statsd.c
+++ b/src/modules/statsd.c
@@ -65,12 +65,25 @@ typedef struct statsd_mod_config {
   int payload_len;
   int ipv4_fd;
   int ipv6_fd;
+  eventer_jobq_t *jobq;
 } statsd_mod_config_t;
 
 typedef struct {
   noit_module_t *self;
   int stats_count;
 } statsd_closure_t;
+
+typedef union {
+  struct sockaddr_in in;
+  struct sockaddr_in6 in6;
+} addr_t;
+
+typedef struct {
+  char *payload;
+  addr_t addr;
+  socklen_t addrlen;
+  noit_module_t *self;
+} statsd_asynch_closure_t;
 
 static int
 statsd_submit(noit_module_t *self, noit_check_t *check,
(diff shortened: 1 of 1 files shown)


src/modules/statsd.c
```diff
@@ -314,30 +370,17 @@ statsd_handler(eventer_t e, int mask, void *closure,
...
-    nchecks += noit_poller_lookup_by_ip_module("0.0.0.0", self->hdr.name,
-                                               checks+nchecks, MAX_CHECKS-1-nchecks);
-    mtevL(nldeb, "statsd(%d bytes) from '%s' -> %d checks%s\n", (int)len,
-          ip, (int)nchecks, parent ? " + a parent" : "");
-    if(parent) checks[nchecks++] = parent;
-    if(nchecks)
-      statsd_handle_payload(checks, nchecks, conf->payload, len);
+    if (conf->jobq) {
+      statsd_asynch_closure_t *sac = (statsd_asynch_closure_t *)malloc(sizeof(statsd_asynch_closure_t));
+      sac->payload = strdup(conf->payload);
```
pjulien, line 376: Not a string per se, need to use `conf->payload_len - 1`

src/modules/statsd.c
```diff
@@ -65,13 +65,26 @@ typedef struct statsd_mod_config {
...
   int stats_count;
 } statsd_closure_t;
 
+typedef union {
+  struct sockaddr_in in;
+  struct sockaddr_in6 in6;
+} addr_t;
+
+typedef struct {
+  char *payload;
```
pjulien, line 82: not a string, need to keep the payload's length

src/modules/statsd.c
```diff
@@ -285,6 +298,55 @@ statsd_handle_payload(noit_check_t **checks, int nchecks,
...
+  if(*ip) {
+    nchecks = noit_poller_lookup_by_ip_module(ip, self->hdr.name, checks, MAX_CHECKS-1);
+  }
+  nchecks += noit_poller_lookup_by_ip_module("0.0.0.0", self->hdr.name,
+    checks+nchecks, MAX_CHECKS-1-nchecks);
+  if(parent) {
+    checks[nchecks++] = parent;
+  }
+  if(nchecks) {
+    statsd_handle_payload(checks, nchecks, payload, strlen(payload));
```
pjulien, line 329: This calls into `update_check` which does seem to be thread safe.  It seems individual threads would update the same data concurrently

postwait: You say it does semn to be thread safe. I question that assertion. I believe that is incorrect.

PR #683, opened 2021-01-25
Fix memory issues with tag search.

postwait:
* Fix uninitialized graphite regex when no regex is found.
* Fix several leaks.

### DIFF
--- a/src/noit_metric_tag_search.c
+++ b/src/noit_metric_tag_search.c
@@ -75,7 +75,7 @@ noit_tag_search_get_tls_pcre_jit_stack(void *arg)
 
 void
 __attribute__((constructor))
-void_tag_search_init(void) {
+noit_tag_search_init(void) {
   pthread_mutex_lock(&noit_tags_search_tls_jit_init_lock);
   if(!noit_tags_search_tls_jit_init) {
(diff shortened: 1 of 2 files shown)


src/noit_metric_tag_search.c
```diff
@@ -49,6 +49,41 @@ typedef struct noit_var_match_t {
...
+static void
+noit_tag_search_free_tls_pcre_jit_stack(void *stack)
+{
+  pcre_jit_stack_free((pcre_jit_stack *)stack);
+}
+static pcre_jit_stack *
+noit_tag_search_get_tls_pcre_jit_stack(void *arg)
+{
+  (void)arg;
+  if (noit_tags_search_tls_jit_init && noit_tag_search_tls_jit_stack == NULL) {
```
pjulien, line 66: One of these is thread local and the other is not?  Doesn't the initialization need to happen once per thread?

postwait: no...  the initialization just needs to happen once.  This skips attempts to do work if it isn't initialized.

There is a key and a __thread to avoid the cost of the _getspecific call once it is allocated on a thread.

pjulien (approved):

PR #690, opened 2021-02-18
Add a setting to SNMP checks to control the max PDU size

rileyberton:
Certain switches *cough*Extreme Networks*cough* don't allow for the increase of the PDU size
like a Cisco does.  If you try to fetch a large number (>400) OIDs in the same PDU, some of the
OIDs will fail to retrieve.

This adds a setting to snmp check to control the max PDU size so that fetches can be kept under the
cap that the device supports.


pamaddox (approved):
LGTM

PR #695, opened 2021-03-26
CIRC-6377 - Set LMDB Checks In Jobq

pamaddox:
When configuring a check via LMDB, offload the work into a jobq rather than doing it in the
main eventer pool.

### DIFF
--- a/src/noit_check_lmdb.c
+++ b/src/noit_check_lmdb.c
@@ -593,9 +593,45 @@ put_retry:
   return 0;
 }
 
+typedef struct lmdb_set_check_data {
+  xmlNodePtr attr;
+  xmlNodePtr config;
+  uuid_t checkid;
+  int error_code;
+  char *error_string;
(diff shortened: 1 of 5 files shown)


src/noit_check_lmdb.c
```diff
@@ -593,17 +594,161 @@ noit_check_lmdb_configure_check(uuid_t checkid, xmlNodePtr a, xmlNodePtr c, int6
...
+      exists = (!target || (check = noit_poller_lookup_by_name(target, name)) != NULL);
+      if(check) {
+        old_seq = check->config_seq;
+      }
+      if(module) {
+        m = noit_module_lookup(module);
+      }
+      rest_check_free_attrs(target, name, module);
+      if(exists) {
+        SET_ERROR_CODE(409, "target`name already registered");
```
pjulien, line 688:
```suggestion
        SET_ERROR_CODE(409, "target name already registered");
```

pjulien (approved):

PR #696, opened 2021-04-12
Fix entity encoding of check config XML generated.

postwait:
We should be using xmlNewTextChild, not xmlNewChild.


pamaddox (approved):
LGTM

bdunavant:
This is for CIRC-6445

PR #699, opened 2021-06-02
Respect the fixed message len for decoding.

postwait:
The code assumed that the message was null terminated which could
cause a read overrun.  Respect the stated length.


vasuraman11 (approved):
LGTM

PR #702, opened 2021-06-23
Expose parsing of metric names to noit_metric_id_t

postwait:


src/noit_message_decoder.c
```diff
@@ -363,6 +371,14 @@ int noit_message_decoder_parse_line(noit_metric_message_t *message, int has_noit
   return 0;
 }
 
+void noit_metric_id_clear(noit_metric_id_t *id) {
+  free(id->alloc_name);
+  free(id->stream.tags);
+  id->stream.tag_count = 0;
+  free(id->measurement.tags);
```
pjulien, line 378: Considering it's now part of the API and enables the reuse of `noit_metric_id_t`, the 3 freed fields should be nullified.

pjulien (approved):

PR #713, opened 2021-08-18
CIRC-6912 - SMR Fixes

pjulien:
- Fix some misuses of mtev SMR
- Fix a couple of eventer_t leaks
- Fix race condition on de-schedule that leads
  to heap use after free
- typo


vasuraman11 (approved):
LGTM

PR #721, opened 2021-09-03
CIRC-6987 - Fix Some Check Reference Counts

pamaddox:
Fix a few issues with check reference counting.


vasuraman11 (approved):
LGTM although I am not familiar with the code paths that lead to these choices

PR #730, opened 2021-09-14
CIRC-7014 - Add Missing SMR Transactions

pjulien:
Add SMR transactions to Lua API

### DIFF
--- a/src/modules/lua_check.c
+++ b/src/modules/lua_check.c
@@ -192,8 +192,8 @@ noit_lua_module_set_name(lua_State *L) {
     }
   }
   else if(lua_gettop(L) > 1) {
-    luaL_error(L, "wrong number of arguments");
     mtev_memory_end();
+    luaL_error(L, "wrong number of arguments");
   }
 
   lua_pushstring(L, module->hdr.name);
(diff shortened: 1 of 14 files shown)


src/modules/lua_check.c
```diff
@@ -205,6 +207,7 @@ noit_lua_module_set_xml_description(lua_State *L) {
   else if(lua_gettop(L) > 1)
     luaL_error(L, "wrong number of arguments");
```
postwait, line 208: This will longjmp out without ending the SMR transaction.

src/modules/lua_check.c
```diff
@@ -430,6 +443,8 @@ noit_lua_set_metric_f(lua_State *L, mtev_boolean allow_whence,
   int64_t __l = 0;
   uint64_t __L = 0;
 
+  mtev_memory_begin();
+
   if(lua_gettop(L) < 2 || lua_gettop(L) > 4) luaL_error(L, "need 2-4 arguments: <metric_name> <value> [whence_s] [whence_us]");
```
postwait, line 448: We can't error inside an SMR transaction b/c it will leave without closing it.

postwait (changes requested):

src/modules/lua_check.c
pjulien: I thought this would kill the process

postwait: No, it will break the runtime just for this check execution.  the top level lua_resume will return an error.

pjulien: Yes, I see, it does a ``_longjmp``

pamaddox (approved):

PR #754, opened 2022-01-14
Expose broker connections via consul services

postwait:


src/stratcon_jlog_streamer.c
```diff
@@ -98,6 +109,83 @@ struct jlog_stream_stats {
...
+    if(sr) {
+      struct ipport tgt = { .port = 0, .target = "unknown" };
+      void *vtgt = NULL;
+      pthread_mutex_lock(&noit_ip_by_cn_lock);
+      if(mtev_hash_retrieve(&noit_ip_by_cn, strdup(cn), strlen(cn),
+                            &vtgt)) {
+        tgt = *(struct ipport *)vtgt;
+      }
+      pthread_mutex_unlock(&noit_ip_by_cn_lock);
+      consul_start(cn, &tgt);
```
pjulien, line 141: I find it confusing that a function called ``consul_stop`` is calling ``consul_start``

pjulien (approved):

src/stratcon_jlog_streamer.c
postwait: That's fair. I'll add a comment.

postwait: Actually, it's likely we want to add a deregister and split the behavior.  It will require a bit more work.

src/stratcon_jlog_streamer.c
```diff
@@ -111,49 +206,182 @@ mtev_hash_table noit_stats_iep; /* this is persistent */
 mtev_hash_table noit_stats_durable; /* this is persistent */
 mtev_hash_table noit_consul_service;
 
+long WARNING_DELAY_MS = 30 * 1000; // 20 seconds
```
pjulien, line 209: either comment or value is wrong

src/stratcon_jlog_streamer.c
```diff
@@ -111,49 +206,182 @@ mtev_hash_table noit_stats_iep; /* this is persistent */
...
+long WARNING_SESSION_TIME_MS = 10 * 1000; // 15 seconds
```
pjulien, line 210: same

PR #757, opened 2022-03-08
Make tag search arguments arrays grow geometrically.

postwait:
Grow by x2 instread of +1.

### DIFF
--- a/src/noit_metric_tag_search.c
+++ b/src/noit_metric_tag_search.c
@@ -565,11 +565,14 @@ typedef struct noit_metric_tag_match_t {
   noit_var_match_t name;
 } noit_metric_tag_match_t;
 
+#define DEFAULT_CHILDREN_ALLOC 16
+
 typedef struct noit_metric_tag_search_ast_t {
   noit_metric_tag_search_op_t operation;
   union {
     struct {
       int cnt;
+      int nallocd;
       struct noit_metric_tag_search_ast_t **node;
     } args;
     noit_metric_tag_match_t spec;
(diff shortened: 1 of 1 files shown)


src/noit_metric_tag_search.c
```diff
@@ -580,19 +583,20 @@ typedef struct noit_metric_tag_search_ast_t {
...
+noit_metric_tag_search_resize_args(noit_metric_tag_search_ast_t *node, int new_size) {
   assert(node->operation == OP_NOT_ARGS ||
          node->operation == OP_OR_ARGS ||
          node->operation == OP_AND_ARGS);
-  if(node->contents.args.cnt == cnt) return;
-  assert(node->contents.args.cnt <= cnt);
-  noit_metric_tag_search_ast_t **new_nodes = calloc(cnt, sizeof(*new_nodes));
+  if(new_size == 0) new_size = DEFAULT_CHILDREN_ALLOC;
+  if(node->contents.args.nallocd >= new_size) return;
+  assert(node->contents.args.nallocd <= new_size);
```
jonInfinity, line 592: This isn't required, line 591 would have already returned

src/noit_metric_tag_search.c
```diff
@@ -666,9 +670,13 @@ void
...
   assert(node->operation != OP_NOT_ARGS || idx == 0);
-  if(idx == node->contents.args.cnt) noit_metric_tag_search_resize_args(node, idx+1);
-  assert(node->contents.args.cnt > idx);
+  noit_metric_tag_search_resize_args(node, node->contents.args.nallocd*2);
+  assert(node->contents.args.nallocd > idx);
+  noit_metric_tag_search_ast_t *prev = node->contents.args.node[idx];
   node->contents.args.node[idx] = r;
+  assert(node->contents.args.cnt >= idx); // can't insert with gaps.
+  if(node->contents.args.cnt == idx) node->contents.args.cnt++;
+  else noit_metric_tag_search_free(prev);
```
jonInfinity, line 678: What happens to prev if node->contents.args.cnt == idx ? It looks like it will leak memory

postwait: If cnt == idx we're appending to the end of an existing list (and not replacing an existing item).  The only idx's that have items in the are less than cnt.

jonInfinity (approved):
LGTM

PR #758, opened 2022-03-09
Native OpenTelemetry metrics sink (otlphttp)

postwait:


src/modules/otlphttp.cpp
```diff
@@ -0,0 +1,905 @@
...
+  name_builder &add(const opentelemetry::proto::common::v1::KeyValue &f) {
+    materialized = false;
+    if(f.has_value()) {
+      add(f.key(), f.value());
+    } else {
+      tags.push_back(std::make_tuple(f.key(), "", false));
+    }
+    return *this;
+  }
+  name_builder &add(std::string cat, std::string name) {
```
pjulien, line 400:
```suggestion
  name_builder &add(const std::string &cat, const std::string &name) {
```

src/modules/otlphttp.cpp
```diff
@@ -0,0 +1,905 @@
...
+  if(mtev_hash_size(rxc->immediate_metrics) > 1000) {
+    metric_local_batch_flush_immediate(rxc);
+    return;
+  }
+  if(mtev_hash_retrieve(rxc->immediate_metrics, cmetric, cmetric_len, &vm)) {
+    /* collision, just log it out */
+    metric_local_batch_flush_immediate(rxc);
+    return;
+  }
+  //metric_t *m = mtev_memory_safe_malloc_cleanup(sizeof(*m), metric_local_free);
```
pjulien, line 647:
```suggestion
```

src/modules/otlphttp.cpp
```diff
@@ -0,0 +1,905 @@
...
+    if(secret && !strcmp(pats[1], secret)) allowed = mtev_true;
+    if(!allowed && cross_module_reverse_allowed(check, pats[1])) allowed = mtev_true;
+
+    if(!allowed) {
+      error = "secret mismatch";
+      goto error;
+    }
+
+    rxc = new otlphttp_upload_t(check);
+    if(const char *mode_str = mtev_hash_dict_get(check->config, "hist_approx_mode"); mode_str) {
```
pjulien, line 717: I'm glad you're using this but in this case don't need the 3rd form of ``if``, the second form will do

```suggestion
    if(const char *mode_str = mtev_hash_dict_get(check->config, "hist_approx_mode")) {
```

src/modules/otlphttp.cpp
```diff
@@ -0,0 +1,905 @@
...
+
+  rxc = rest_get_upload(restc, &mask, &complete);
+  if(rxc == NULL && !complete) return mask;
+
+  if(!rxc) {
+    error = "No data?";
+    goto error;
+  }
+  
+  if(true) {
```
pjulien, line 750: left-over from debug?

pjulien (approved):

PR #759, opened 2022-03-11
CIRC-7654 Search Hinting On Match Nodes

pamaddox:
Allow setting various search hints on tag search nodes. Consumers can then utilize these hints to alter search behavior.


postwait (commented):
reconnoiter doesn't know anything about indexes.  the hint arguments shouldn't be limited to index:none etc.  It should allow any cat:name tag name.

pamaddox:
Going to go in a different direction with this

PR #761, opened 2022-03-14
CIRC-7654 - Make Const Things Const In Tag Search

pamaddox:
These fields are not modified, so we should flag them as const.


vasuraman11 (approved):
LGTM

PR #766, opened 2022-03-18
prometheus: return 4xx codes on bad client input

postwait:


bdunavant (approved):
I wholeheartedly approve.  LGTM

PR #773, opened 2022-04-08
Add Tag Search Swap Function

pamaddox:
Added a function to noit_metric_tag_search that will allow taking any
two elements in a list and having them swap places.

### DIFF
--- a/src/noit_metric_tag_search.c
+++ b/src/noit_metric_tag_search.c
@@ -1255,3 +1255,34 @@ noit_metric_tag_search_unparse(const noit_metric_tag_search_ast_t *search) {
   mtev_dyn_buffer_destroy(&buf);
   return res;
 }
+
+int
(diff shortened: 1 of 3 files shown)


src/noit_metric_tag_search.c
```diff
@@ -1255,3 +1255,34 @@ noit_metric_tag_search_unparse(const noit_metric_tag_search_ast_t *search) {
...
+  if (!search) {
+    return -1;
+  }
+  if (idx1 < 0 || idx2 < 0) {
+    return -1;
+  }
+  int nargs = noit_metric_tag_search_get_nargs(search);
+  if (nargs == 0) {
+    return -1;
+  }
```
postwait, line 1270: This == 0 is redundant to the next predicate

postwait (approved):
Nit

src/noit_metric_tag_search.c
pamaddox: Fixed

postwait (approved):

PR #775, opened 2022-04-18
Make invlid regular expressions safe.

postwait:
tag search 're' expressions that don't compile should not crash the
matcher.


mcieslinski (approved):
Seems okay

PR #777, opened 2022-04-28
Add Filterrule Type To Apply Measurement Tags

pamaddox:
Add a new filterrule type, `add_measurement_tag:<tag>:<cat>`, that will apply a measurement tag and move on to the next rule in the filterset if it is encountered.

### DIFF
--- a/src/noit_filters.c
+++ b/src/noit_filters.c
@@ -819,6 +819,10 @@ noit_apply_filterset(const char *filterset,
           skipto_rule = r->skipto_rule;
           continue;
         }
+        else if (r->type == NOIT_FILTER_ADD_MEASUREMENT_TAG) {
+          noit_add_measurement_tag(r, &mtset);
+          continue;
+        }
         ck_pr_inc_32(&r->matches);
         ret = (r->type == NOIT_FILTER_ACCEPT) ? mtev_true : mtev_false;
         break;
--- a/src/noit_filters.c
+++ b/src/noit_filters.c
@@ -768,7 +768,11 @@ noit_apply_filterset(const char *filterset,
           continue;
         }
         else if (r->type == NOIT_FILTER_ADD_MEASUREMENT_TAG) {
-          noit_add_measurement_tag(r, &mtset);
+          if (noit_add_measurement_tag(r, &mtset)) {
+            mtevL(nf_error, "could not apply measurement tag - <%s:%s>\n",
+              r->measurement_tag.add_measurement_tag_cat,
+              r->measurement_tag.add_measurement_tag_val ? r->measurement_tag.add_measurement_tag_val : "");
+          }
           continue;
         }
         ck_pr_inc_32(&r->matches);
@@ -820,7 +824,11 @@ noit_apply_filterset(const char *filterset,
           continue;
         }
         else if (r->type == NOIT_FILTER_ADD_MEASUREMENT_TAG) {
-          noit_add_measurement_tag(r, &mtset);
+          if (noit_add_measurement_tag(r, &mtset)) {
+            mtevL(nf_error, "could not apply measurement tag - <%s:%s>\n",
+              r->measurement_tag.add_measurement_tag_cat,
+              r->measurement_tag.add_measurement_tag_val ? r->measurement_tag.add_measurement_tag_val : "");
+          }
           continue;
         }
         ck_pr_inc_32(&r->matches);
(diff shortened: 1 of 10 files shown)


src/noit_filters.c
```diff
@@ -328,6 +339,25 @@ noit_filter_compile_add(mtev_conf_section_t setinfo) {
       rule->type = NOIT_FILTER_SKIPTO;
       rule->skipto = strdup(buffer+strlen(FILTERSET_SKIPTO_STRING));
     }
+    else if (!strncasecmp(buffer, FILTERSET_ADD_MEASUREMENT_TAG_STRING, strlen(FILTERSET_ADD_MEASUREMENT_TAG_STRING))) {
```
vasuraman11, line 342: here you are not case sensitive, but in other places it uses strncmp, is this intentional?

src/noit_filters.c
```diff
@@ -655,6 +685,76 @@ noit_apply_filterrule_metric(filterrule_t *r,
...
+                         char **expanded_metric_name,
+                         noit_metric_tagset_t *mtset) {
+  if (!r || !metric || !expanded_metric_name || !mtset || !r->measurement_tag.add_measurement_tag_cat || mtset->tag_count >= MAX_TAGS) {
+    return -1;
+  }
+  char encoded_nametag[NOIT_TAG_MAX_PAIR_LEN+1];
+  char decoded_nametag[NOIT_TAG_MAX_PAIR_LEN+1];
+  const char *cat = r->measurement_tag.add_measurement_tag_cat;
+  const char *val = r->measurement_tag.add_measurement_tag_val ? r->measurement_tag.add_measurement_tag_val : "";
+  snprintf(decoded_nametag, sizeof(decoded_nametag), "%s%c%s", cat, NOIT_TAG_DECODED_SEPARATOR, val);
```
vasuraman11, line 700: what if there is a cat but empty val?  does this still work?

vasuraman11 (commented):
see comments, looks pretty good

src/noit_filters.c
```diff
@@ -328,6 +339,25 @@ noit_filter_compile_add(mtev_conf_section_t setinfo) {
...
+      rule->type = NOIT_FILTER_ADD_MEASUREMENT_TAG;
+      rule->measurement_tag.add_measurement_tag_cat = NULL;
+      rule->measurement_tag.add_measurement_tag_val = NULL;
+      const char *mt = buffer+strlen(FILTERSET_ADD_MEASUREMENT_TAG_STRING);
+      const char *colon = strstr(mt, ":");
+      if (colon) {
+        size_t cat_len = colon - mt;
+        if (cat_len) {
+          rule->measurement_tag.add_measurement_tag_cat = (char *)calloc(1, cat_len + 1);
+          memcpy(rule->measurement_tag.add_measurement_tag_cat, mt, cat_len);
```
pjulien, line 352: ``add_measurement_tag_cat`` is null terminated on line 353 but not here

src/noit_filters.c
```diff
@@ -655,6 +685,76 @@ noit_apply_filterrule_metric(filterrule_t *r,
...
+noit_update_metric_name(char *expanded_metric_name,
+                        noit_metric_tagset_t *mtset,
+                        int mtset_tag_start,
+                        metric_t *metric) {
+  char buff[MAX_METRIC_TAGGED_NAME];
+  char *old = metric->metric_name;
+  if (!expanded_metric_name) {
+    expanded_metric_name = strdup(old);
+  }
+  if (noit_metric_canonicalize(expanded_metric_name, strlen(expanded_metric_name), buff, sizeof(buff), mtev_true)) {
```
pjulien, line 743: Can you check ``noit_metric_canonicalize_ex`` while you're here?  I think it can overflow its output argument by 1 byte.


```c
  if(null_term) output[out-buff] = '\0';
  if((out-buff) > MAX_METRIC_TAGGED_NAME) return -8;
```

src/noit_filters.c
pamaddox: Cat with empty val is valid, val with empty cat is not

src/noit_filters.c
pamaddox: I'm matching the style of what's already here. If we change the style on the new code, we should probably change it on everything, which would theoretically be done in another PR

src/noit_filters.c
pamaddox: It is null terminated. We use `calloc`, which zero-fills the memory. We allocate `cat_len + 1`, then memcpy `cat_len`. That results in an extra allocated byte at the end that is null.

bosley (approved):
Looks solid :fist:

PR #779, opened 2022-05-11
HTTPTrap And Histogram Measurement Tag Fixes

pamaddox:
Two quick changes:
* Fix httptrap module to use newly allocated memory after applying
  filtersets
* Don't apply measurement tags for histograms or METRIC_GUESS
  metrics - these are tricky and need more care before they're ready.

### DIFF
--- a/src/modules/httptrap.c
+++ b/src/modules/httptrap.c
@@ -139,10 +139,14 @@ struct rest_json_payload {
 };
 
 static void
-track_filtered(struct rest_json_payload *json, const char *name) {
-  metric_t m = { .metric_name = (char *)name };
+track_filtered(struct rest_json_payload *json, char **name) {
+  metric_t m = { .metric_name = *name };
+  char *before = m.metric_name;
   if(!noit_apply_filterset(json->check->filterset, json->check, &m)) {
     json->filtered_cnt++;
+  }
+  if (before != m.metric_name) {
+    *name = m.metric_name;
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/modules/httptrap.c
```diff
@@ -533,7 +537,7 @@ httptrap_yajl_cb_end_map(void *ctx) {
        * in progress metrics only (get_metric) not (get_last_metric)
        * and we also much fetch a count...
        */
-      track_filtered(json, metric_name);
+      track_filtered(json, &metric_name);
```
pjulien, line 540: who frees ``metric_name`` if the value was changed?

src/modules/httptrap.c
```diff
@@ -314,7 +318,7 @@ httptrap_yajl_cb_null(void *ctx) {
   if(rv) return 1;
   if(json->keys[json->depth]) {
     _YD("[%3d] cb_null\n", json->depth);
-    track_filtered(json, json->keys[json->depth]);
+    track_filtered(json, &json->keys[json->depth]);
```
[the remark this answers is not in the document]
pamaddox: No it doesn't. track_filtered calls `noit_apply_filterset`, which will free the previous key if the key changes.

src/modules/httptrap.c
pamaddox: `track_filtered` calls `noit_apply_filterset`, which will free it.

pjulien (approved):

PR #783, opened 2022-05-19
Add Debug Information When Message Decode Fails

pamaddox:
Replace an assert with a mtevFatal that gives details about the cause of
the failure.


vasuraman11 (approved):
LGTM

pjulien (approved):

PR #789, opened 2022-07-14
CIRC-8763 - Divide by zero

pjulien:
https://circonus.sp.backtrace.io/p/broker/debug?time=all&fingerprint=feeab3ecdbb16668b93883cce6e688ecd2a82d0475e8f7521307c67f3ffdfc5e&debug=(%22533ec%22,0,5,24)

### DIFF
--- a/src/modules/histogram.c
+++ b/src/modules/histogram.c
@@ -291,6 +291,7 @@ static void
 update_histotier(histotier *ht, mtev_boolean cumulative, uint64_t s,
                  struct histogram_config *conf, noit_check_t *check,
                  const char *name, double val, uint64_t cnt, const histogram_t *hist) {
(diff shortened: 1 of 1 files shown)


src/modules/histogram.c
```diff
@@ -291,6 +291,7 @@ static void
 update_histotier(histotier *ht, mtev_boolean cumulative, uint64_t s,
                  struct histogram_config *conf, noit_check_t *check,
                  const char *name, double val, uint64_t cnt, const histogram_t *hist) {
+  if (s/ht->cadence == 0) return;
```
pamaddox, line 294: What if `ht->cadence` is zero? Is that something we should check for as well?

postwait (approved):

PR #790, opened 2022-07-18
CIRC-8777 - use after free

pjulien:
When using ``noit_stats_mark_metric_logged`` it takes ownership of the passed
in ``metric_t`` pointer.  This creates a race between the 3 updated call sites
clearing, and freeing, their hash table versus the one found inside
``noit_stats``

Harmonize everything with ownership under ``noit_stats`` and the call sites
respecting that

- Remove call to ``noit_check_get_stats_current``, return value is not used
- Consolidate ``metric_local_free`` code
- Only use SMR to allocate ``metric_t``
- ``immediate_metrics`` does not own anything, do not delete from there
- Add a couple of ``default`` clauses, Eclipse paints the entire switch
  statement in yellow otherwise

### DIFF
--- a/src/noit_socket_listener.c
+++ b/src/noit_socket_listener.c
@@ -60,15 +60,6 @@
 static mtev_log_stream_t nldeb = NULL;
 static mtev_log_stream_t nlerr = NULL;
 
-static void metric_local_free(void *vm) {
-  if(vm) {
-    metric_t *m = vm;
-    free(m->metric_name);
-    free(m->metric_value.vp);
-  }
-  free(vm);
-}
-
 /* Count endlines to determine how many full records we
  * have */
 static inline int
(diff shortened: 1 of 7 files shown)


src/noit_socket_listener.c
```diff
@@ -254,12 +245,16 @@ listener_metric_track_or_log(void *vrxc, const char *name,
       }
     }
   }
-  noit_stats_mark_metric_logged(noit_check_get_stats_inprogress(rxc->check), m, mtev_false);
-  pthread_mutex_lock(&rxc->flushlock);
-  mtev_boolean inserted = mtev_hash_store(rxc->immediate_metrics, m->metric_name, strlen(m->metric_name), m);
```
postwait, line 259: I think this changes the logic in a way that will cause data loss.

postwait (approved):
These changes appear to work for me.

src/noit_socket_listener.c
pjulien: how?  If it's not in ``logged`` how can it be in ``immediate_metrics``?

postwait: I was worried about it being logged and then showing up immediately again where it would find itself already logged and skip the immediate insert.

PR #791, opened 2022-07-25
Add Tag Compaction Hook

pamaddox:
Added a hook into the metric tag compaction code to allow user-defined
functions to specify tags to remove from the set.

### DIFF
--- a/src/noit_message_decoder.h
+++ b/src/noit_message_decoder.h
@@ -41,6 +41,11 @@
 extern "C"  {
 #endif
 
+MTEV_HOOK_PROTO(noit_message_decoder_compact_validate_tag,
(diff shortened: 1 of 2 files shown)


postwait (changes requested):
This is atypical hook design because you are aiming a different meaning for ABORT.

The hook should have a bool * param that is checked to see if the tag is acceptable

src/noit_message_decoder.h
```diff
@@ -41,6 +41,12 @@
 extern "C"  {
 #endif
```
pjulien, line 42: should include ``stdbool.h`` since this is C

pamaddox:
This is too unconditional; gonna do this with callbacks instead.

PR #794, opened 2022-07-25
Add Validation Callback To Tagset Compaction

pamaddox:
Add a callback function (wrapped in a context struct for ease of later
expansion) to allow passing a custom validation function to drop certain
tags when doing tag compaction.

### DIFF
--- a/src/noit_metric.c
+++ b/src/noit_metric.c
@@ -859,6 +859,13 @@ noit_metric_tagset_context_free(noit_metric_tagset_context_t *ctx) {
 }
 
 void
+noit_metric_tagset_context_ref(noit_metric_tagset_context_t *ctx) {
+  if (ctx) {
+    ck_pr_inc_32(&ctx->refcnt);
+  }
+}
+
+void
 noit_metric_tagset_context_set_validate_function(noit_metric_tagset_context_t *ctx,
                                                  bool (*f)(const char * cat,
                                                            const size_t cat_len,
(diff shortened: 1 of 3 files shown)


src/noit_metric.c
```diff
@@ -836,3 +842,41 @@ const char *
...
+noit_metric_tagset_context_alloc(void) {
+  noit_metric_tagset_context_t *ctx = (noit_metric_tagset_context_t *)malloc(sizeof(*ctx));
+  ctx->refcnt = 1;
+  ctx->validate_function = NULL;
+  return ctx;
+}
+
+void
+noit_metric_tagset_context_free(noit_metric_tagset_context_t *ctx) {
+  if (ck_pr_dec_32_is_zero(&ctx->refcnt)) {
```
pjulien, line 856: why do we have a ``refcnt`` if we never increment it?

postwait: We should have a _ref() function and we need to ref it when it is attached to the operation

pamaddox: If I understand correctly: The correct usage of this would be:

alloc (set ref count to 1)
attach validation function (increments ref count to 2)
execute validation function
free (reduces coult to 1)
free (reduces count to 0, actually frees)

pamaddox: added a ref function

pjulien (approved):

PR #798, opened 2022-08-08
Use Generations To Determine If Resizes Are Needed

pamaddox:
Rather than checking to see if we're using at least 80% of the LMDB
map, use generations to determine whether another thread has resized or
not.

### DIFF
--- a/src/noit_lmdb_tools.c
+++ b/src/noit_lmdb_tools.c
@@ -335,16 +335,24 @@ void noit_lmdb_tools_close_instance(noit_lmdb_instance_t *instance)
   free(instance);
 }
 
+uint64_t noit_lmdb_get_instance_generation(noit_lmdb_instance_t *instance)
+{
+  return ck_pr_load_64(&instance->generation);
+}
+
+static void noit_lmdb_increment_instance_generation(noit_lmdb_instance_t *instance)
+{
+  ck_pr_inc_64(&instance->generation);
+}
+
 #define NOIT_LMDB_RESIZE_FACTOR 1.5
-void noit_lmdb_resize_instance(noit_lmdb_instance_t *instance)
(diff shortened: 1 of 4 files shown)


src/noit_lmdb_tools.h
```diff
@@ -71,6 +72,8 @@ void noit_lmdb_free_filterset_data(noit_lmdb_filterset_rule_data_t *data);
 noit_lmdb_instance_t *noit_lmdb_tools_open_instance(char *path);
 void noit_lmdb_tools_close_instance(noit_lmdb_instance_t *instance);
 void noit_lmdb_resize_instance(noit_lmdb_instance_t *instance);
+uint64_t noit_lmdb_get_instance_generation(noit_lmdb_instance_t *instance);
+void noit_lmdb_increment_instance_generation(noit_lmdb_instance_t *instance);
```
postwait, line 76: This doesn't need to be accessible

src/noit_lmdb_tools.c
```diff
@@ -342,16 +343,16 @@ void noit_lmdb_resize_instance(noit_lmdb_instance_t *instance)
   uint64_t new_mapsize;
 
   /* prevent new transactions on the write side */
+
+  uint64_t initial_generation = noit_lmdb_get_instance_generation(instance);
```
postwait, line 347: It isn't safe to call this here. This `initial_generation` should be an argument to the resize function.  The called should call the `noit_lmdb_get_isntance_generation(instance);` upon getting a MAP_FULL before releasing the lock for resize

postwait (changes requested):

postwait (approved):

pjulien (approved):
I do wish however that all ``initial_generation`` variables were ``const`` though

pjulien (approved):

PR #799, opened 2022-08-09
LMDB Tweaks

pamaddox:
* Use a starting map size aligned with pagesize that is at least 64MB
* Resize DBs by a factor of 4
* Improve resize print messages

### DIFF
--- a/src/noit_lmdb_tools.c
+++ b/src/noit_lmdb_tools.c
@@ -360,7 +360,7 @@ static void noit_lmdb_increment_instance_generation(noit_lmdb_instance_t *instan
   ck_pr_inc_64(&instance->generation);
 }
 
-#define NOIT_LMDB_RESIZE_FACTOR 4
+#define NOIT_LMDB_RESIZE_FACTOR 2
 void noit_lmdb_resize_instance(noit_lmdb_instance_t *instance, const uint64_t initial_generation)
 {
   MDB_envinfo mei;
--- a/src/noit_lmdb_tools.c
+++ b/src/noit_lmdb_tools.c
@@ -284,15 +284,8 @@ noit_lmdb_instance_t *noit_lmdb_tools_open_instance(char *path)
     return NULL;
   }
 
-  /* Align mapsize to pagesize, use enough pages to reach
-   * 64 MB */
-  size_t page_size = getpagesize();
-  size_t map_size = page_size;
-  static const size_t map_size_target = (1024*1024*64);
-  while (map_size < map_size_target) {
-    map_size += page_size;
-  }
-  rc = mdb_env_set_mapsize(env, map_size);
+  /* Set initial pagesize to 64 MiB */
+  rc = mdb_env_set_mapsize(env, 1024*1024*64);
   if (rc != 0) {
     errno = rc;
     mdb_env_close(env);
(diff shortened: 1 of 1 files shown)


src/noit_lmdb_tools.c
```diff
@@ -345,7 +360,7 @@ static void noit_lmdb_increment_instance_generation(noit_lmdb_instance_t *instan
   ck_pr_inc_64(&instance->generation);
 }
 
-#define NOIT_LMDB_RESIZE_FACTOR 1.5
+#define NOIT_LMDB_RESIZE_FACTOR 4
```
esproul, line 363: Since we are starting larger, we can probably use a factor of 2 instead of 4.

src/noit_lmdb_tools.c
```diff
@@ -284,6 +284,21 @@ noit_lmdb_instance_t *noit_lmdb_tools_open_instance(char *path)
...
 
+  /* Align mapsize to pagesize, use enough pages to reach
+   * 64 MB */
+  size_t page_size = getpagesize();
+  size_t map_size = page_size;
+  static const size_t map_size_target = (1024*1024*64);
+  while (map_size < map_size_target) {
+    map_size += page_size;
+  }
+  rc = mdb_env_set_mapsize(env, map_size);
```
pjulien, line 295: can just set this to 64 MiB here directly

pamaddox: If we do that, how do we know that aligns with the local pagesize?

postwait: map_size = page_size * ((map_size_target + (page_size-1)) / page_size);

pjulien: You don't need to do that yourself, lmdb will align the size if it doesn't match the page size.

However if it wasn't, we know the page size would be a power of 2

postwait (changes requested):

src/noit_lmdb_tools.c
```diff
@@ -284,6 +284,21 @@ noit_lmdb_instance_t *noit_lmdb_tools_open_instance(char *path)
     return NULL;
   }
 
+  /* Align mapsize to pagesize, use enough pages to reach
+   * 64 MB */
+  size_t page_size = getpagesize();
+  size_t map_size = page_size;
+  static const size_t map_size_target = (1024*1024*64);
```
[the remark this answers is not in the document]
pamaddox: Removed variable entirely.

src/noit_lmdb_tools.c
pamaddox: I used Patrick's suggestion, since these two comments conflict. Can change to something else if needed.

pjulien (approved):

src/noit_lmdb_tools.c
```diff
@@ -284,6 +284,14 @@ noit_lmdb_instance_t *noit_lmdb_tools_open_instance(char *path)
     return NULL;
   }
 
+  /* Set initial pagesize to 64 MiB */
```
esproul, line 287: This is the map size, not the page size, correct?

pamaddox: Yup, typo. Fixed.

postwait (approved):

PR #803, opened 2022-08-17
compile and study search AST on-demand.

postwait:
Add parse_lazy and clone_lazy functions to avoid PCRE compilation during
parsing.  Do not study PCREs by default, track executions and once we
have 50 calls to pcre_exec then study the PCRE.  Rework the lfu such
that the token is tracked with the returned object and is released once
the node is freed.

CIRC-8958

### DIFF
--- a/src/noit_metric_tag_search.c
+++ b/src/noit_metric_tag_search.c
@@ -77,6 +77,8 @@ static pthread_key_t tls_state_key;
 
 static bool re_matcher_compile(re_matcher_t *rem)
 {
+  ck_pr_fence_load();
+  if(rem->re || rem->error_offset >= 0) return true;
   pcre *re = NULL;
   if((re = ck_pr_load_ptr(&rem->re)) == NULL && rem->error_offset < 0) {
     const char *error;
@@ -90,6 +92,8 @@ static bool re_matcher_compile(re_matcher_t *rem)
 }
 static inline bool re_matcher_study(re_matcher_t *rem)
 {
+  ck_pr_fence_load();
+  if(rem->re_e || rem->studied) return true;
   pcre_extra *re_e = ck_pr_load_ptr(&rem->re_e);
   if(!rem->studied && re_e == NULL && rem->re != NULL) {
     const char *error;
--- a/src/noit_metric_tag_search.c
+++ b/src/noit_metric_tag_search.c
@@ -66,14 +66,13 @@ typedef struct re_impl {
 } re_impl_t;
 
 typedef struct tls_state {
-  pcre_jit_stack *jit_stack;
   mtev_lfu_t *pattern_to_re_matcher;
 } tls_state_t;
 
 static mtev_boolean noit_match_str(const char *subj, int subj_len, const struct noit_var_match_t *m);
 static tls_state_t *tls_state_alloc(void);
 static pcre_jit_stack *tls_state_get_pcre_jit_stack(void *arg);
-static pthread_key_t tls_state_key;
+static pthread_key_t tls_state_key, jit_stack_key;
 
 static bool re_matcher_compile(re_matcher_t *rem)
 {
@@ -160,7 +159,6 @@ tls_state_free(void *v)
 {
   tls_state_t *st = v;
 
-  pcre_jit_stack_free(st->jit_stack);
   mtev_lfu_destroy(st->pattern_to_re_matcher);
   free(st);
 }
@@ -172,7 +170,6 @@ tls_state_alloc(void)
 
   if (st == NULL) {
     st = calloc(1, sizeof(*st));
-    st->jit_stack = pcre_jit_stack_alloc(1024 * 32, 512 * 1024);
     st->pattern_to_re_matcher = mtev_lfu_create(100000, re_matcher_free);
     pthread_setspecific(tls_state_key, st);
   }
@@ -180,10 +177,22 @@ tls_state_alloc(void)
   return st;
 }
 
+static void tls_state_free_pcre_jit_stack(void *arg)
+{
+  if (arg != NULL) {
+    pcre_jit_stack_free((pcre_jit_stack *)arg);
+  }
+}
+
 static pcre_jit_stack *tls_state_get_pcre_jit_stack(void *arg)
 {
   (void)arg;
-  return tls_state_alloc()->jit_stack;
+  pcre_jit_stack *jit_stack = pthread_getspecific(jit_stack_key);
+  if (jit_stack == NULL) {
+    jit_stack = pcre_jit_stack_alloc(1024 * 32, 512 * 1024);
+    pthread_setspecific(jit_stack_key, jit_stack);
+  }
+  return jit_stack;
 }
 
 static re_matcher_t *
@@ -215,6 +224,7 @@ noit_tag_search_init(void) {
 
   if (!tls_state_key_init) {
     pthread_key_create(&tls_state_key, tls_state_free);
+    pthread_key_create(&jit_stack_key, tls_state_free_pcre_jit_stack);
     tls_state_key_init = true;
   }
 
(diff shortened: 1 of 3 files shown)


src/noit_metric_tag_search.c
```diff
@@ -65,21 +75,58 @@ static tls_state_t *tls_state_alloc(void);
...
+static bool re_matcher_compile(re_matcher_t *rem)
 {
-  const char *error;
-  re_matcher_t *rem = malloc(sizeof(*rem));
-
-  rem->re_str = pattern;
-  rem->re = pcre_compile(pattern, 0, &error, &rem->error_offset, NULL);
-
+  pcre *re = NULL;
+  if((re = ck_pr_load_ptr(&rem->re)) == NULL && rem->error_offset < 0) {
```
pjulien, line 81: no need for atomics, the cache is thread local

src/noit_metric_tag_search.c
```diff
@@ -65,21 +75,58 @@ static tls_state_t *tls_state_alloc(void);
...
+    if(!ck_pr_cas_ptr(&rem->re, NULL, re)) {
+      pcre_free(re);
+      re = ck_pr_load_ptr(&rem->re);
+    }
+  }
+  return re != NULL;
+}
+static inline bool re_matcher_study(re_matcher_t *rem)
+{
+  pcre_extra *re_e = ck_pr_load_ptr(&rem->re_e);
```
pjulien, line 93: no need for atomics, the cache is thread local, it has to be for the jit stack to work properly too

src/noit_metric_tag_search.c
```diff
@@ -136,24 +183,21 @@ static pcre_jit_stack *tls_state_get_pcre_jit_stack(void *arg)
...
-    free(pattern);
-  }
-  else {
+  if (!rem) {
     rem = re_matcher_alloc(pattern);
     mtev_lfu_put(lfu, pattern, pattern_len, rem);
+    *token = mtev_lfu_get(lfu, pattern, pattern_len, (void **) &rem);
   }
-
-  mtev_lfu_release(lfu, token);
```
pjulien, line 156: This was fine, the cache is thread local, but ok

src/noit_metric_tag_search.c
```diff
@@ -214,6 +262,7 @@ void var_graphite_free(void *vi) {
   struct graphite_impl *g = vi;
   if(g == NULL) return;
   for(int i=0; i<g->count; i++) {
+    mtev_lfu_release_f(re_matcher_free, g->results[i].tok);
```
pjulien, line 265: not needed, ``re_matcher_free`` was already set as the "free" function when the cache was created

```suggestion
    mtev_lfu_release((g->results[i].tok);
```

src/noit_metric_tag_search.c
```diff
@@ -486,38 +554,51 @@ build_regex_from_expansion(const char *expansion) {
...
   (void)pattern;
-  struct re_matcher *m = (struct re_matcher *)impl_data;
+  re_impl_t *m = (re_impl_t *)impl_data;
   if(m == NULL) return mtev_false;
   int ovector[30], rv;
-  rv = pcre_exec(m->re, m->re_e, in, in_len, 0, 0, ovector, 30);
+  if(m->rem->re == NULL) {
+    re_matcher_compile(m->rem);
+  }
+  ck_pr_faa_64(&m->rem->execution_counter, 1);
```
pjulien, line 586: don't need atomic

src/noit_metric_tag_search.c
```diff
@@ -486,38 +554,51 @@ build_regex_from_expansion(const char *expansion) {
...
+  rv = pcre_exec(m->rem->re, m->rem->re_e, in, in_len, 0, 0, ovector, 30);
   if(rv >= 0) return mtev_true;
   return mtev_false;
 }
 
-static void no_op_re_free(void *impl_data) {
-  // no-op, resides in LFU cache
+static void var_re_free(void *impl_data) {
+  re_impl_t *m = (re_impl_t *)impl_data;
+  mtev_lfu_release_f(re_matcher_free, m->tok);
```
pjulien, line 595:
```suggestion
  mtev_lfu_release(m->tok);
```

src/noit_metric_tag_search.c
```diff
@@ -714,7 +799,7 @@ noit_metric_tag_search_add_arg(noit_metric_tag_search_ast_t *node, noit_metric_t
 
 noit_metric_tag_search_ast_t *
 noit_metric_tag_search_alloc(noit_metric_tag_search_op_t op) {
-  noit_metric_tag_search_ast_t *node = calloc(1, sizeof(*node));
+  noit_metric_tag_search_ast_t *node = calloc(sizeof(*node), 1);
```
pjulien, line 802: this was correct per the manual page

src/noit_metric_tag_search.c
```diff
@@ -818,7 +915,7 @@ noit_metric_tag_match_compile(struct noit_var_match_t *m, const char **endq, int
     if(**endq != ']') {
       *endq = query+1;
       return mtev_false;
-    }
+    }
```
pjulien, line 918:
```suggestion
    }
```

src/noit_metric_tag_search.c
```diff
@@ -200,8 +243,13 @@ var_graphite_match(void *impl_data, const char *pattern, const char *in, size_t
   /* If the fixed_prefix len is the subject len, we know we have an exact match */
   if(g->fixed_prefix_len == in_len) return mtev_true;
   for(int i=0; i<g->count; i++) {
-    if(g->results[i].re) {
-      rv = pcre_exec(g->results[i].re, g->results[i].re_e, in, in_len, 0, 0, ovector, 30);
+    if(g->results[i].rem && g->results[i].rem->re == NULL) {
+      re_matcher_compile(g->results[i].rem);
+    }
+    if(g->results[i].rem && g->results[i].rem->re) {
+      ck_pr_faa_64(&g->results[i].rem->execution_counter, 1);
```
pjulien, line 250:
```suggestion
      g->results[i].rem->execution_counter++;
```

src/noit_metric_tag_search.c
postwait: The cache is thread local, the item in the cache is not necessarily single thread.  It can (and is) passed to other threads.  Now that the re is compiled on demand, we need to use atomics to not race/leak on the compile and assignment.

src/noit_metric_tag_search.c
postwait: It doesn't need to be thread local for the jit stack to work, that's why we pass in a function to create the stack and not just pass in memory.  This tls jit stack factory is used elsewhere as well.

src/noit_metric_tag_search.c
postwait: The cache is thread local, the rem is not guaranteed to be accessed/executed from a single thread.

src/noit_metric_tag_search.c
postwait: lfu_release take the LFU itself as the first argument... which could have been destroyed by the time the thread using this AST ultimaltely releases it.

src/noit_metric_tag_search.c
pjulien: How can this possibly work?  Once the stack is assigned to a regex, it's still the same shared stack for all the assignments that were done on this thread.

If the regex is moved with its stack to another thread, then it will be used by multiple threads at the same time.

The study should only be done when requesting a regex from this API, at the very least this would limit the atomics to only the execution counter

postwait: b/c the stack isn't assigned to the regex.  The function to get the stack is called as needed.  And it returns a thread local stack each time.

pjulien: checking the source of pcre_assign_jit_stack to see if the function is called right away or when executing and what it's assigned to

pjulien: > b/c the stack isn't assigned to the regex. The function to get the stack is called as needed. And it returns a thread local stack each time.

OK then, but this should still be a single atomic for the count.  We get it from the cache and we check its execution counter and swap it in

pjulien (approved):

pjulien (approved):

pjulien (approved):

pjulien (approved):

PR #807, opened 2022-08-29
Add Size-Limited Noit Metric Find Unparse Function

pamaddox:
Add noit_metric_tag_search_unparse_limit_size, which allows setting a
size limit to ensure that we don't go overboard with extremely large
queries.

Limit recursion to make sure we don't blow out the stack.

### DIFF
--- a/src/noit_metric_tag_search.c
+++ b/src/noit_metric_tag_search.c
@@ -43,6 +43,7 @@
 #include <ctype.h>
 
 int STUDY_EXECUTION_THRESHOLD = 50;
+int64_t NOIT_METRIC_TAG_SEARCH_UNPARSE_RECURSION_LIMIT = 1000;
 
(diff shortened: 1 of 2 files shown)


src/noit_metric_tag_search.c
```diff
@@ -43,6 +43,7 @@
 #include <ctype.h>
 
 int STUDY_EXECUTION_THRESHOLD = 50;
+int64_t NOIT_METRIC_TAG_SEARCH_UNPARSE_RECURSION_LIMIT = 1000;
```
pjulien, line 46:
```suggestion
const int64_t NOIT_METRIC_TAG_SEARCH_UNPARSE_RECURSION_LIMIT = 1000;
```

pjulien (approved):

PR #811, opened 2022-10-07
Update build instructions

esproul:
Much has changed since they were last updated. There are many more dependencies, libmtev has been carved out, and the OSes have evolved.


bdunavant (approved):
lgtm

PR #812, opened 2022-10-21
CIRC-9320 - heap use after free

pjulien:
Protect the hash table's contents using a seperate lock.

Although the mtev hash table itself has internal syncrhonization, those do not protect the contents that are stored in the hash table itself by the user, e.g.,

mtev_hash_replace(r->rname##_ht, strdup(value), strlen(value), NULL, free, NULL); \

The value stored at this location is made using strdup.  Someone can still call free on the value and have that value still be visible from other threads


pamaddox (approved):

vasuraman11 (approved):
LGTM

PR #815, opened 2022-11-10
Move Check Recycler Into A Jobq

pamaddox:
This process can take a long time and is currently on the main thread, which can cause things to block or watchdog. It is not time time-sensitive task, so we should move it into a jobq to keep the main thread clear.


vasuraman11 (approved):
LGTM

PR #816, opened 2022-11-15
Move Filter Set Endpoint Into Jobq

pamaddox:


src/noit_filters_lmdb.c
```diff
@@ -1569,39 +1664,33 @@ noit_filters_lmdb_rest_set_filter(mtev_http_rest_closure_t *restc,
...
+   * it when we clean up */
+  lfsc->xml_data = restc->call_closure;
+  lfsc->xml_data_free = restc->call_closure_free;
+
+  restc->call_closure = lfsc;
+  restc->call_closure_free = lmdb_set_filter_data_free;
+  restc->fastpath = noit_filters_lmdb_rest_set_filter_complete;
+
+  eventer_t e = eventer_alloc_asynch(noit_filters_lmdb_rest_set_filter_asynch, lfsc);
+  if(conne) eventer_set_owner(e, eventer_get_owner(conne));
```
pjulien, line 1689: I don't believe this is sufficient to only set the owner by the HTTP connection.   

```c++
noit_filters_lmdb_already_in_db(lfsc->name);
```

Calls like this one could still produce incorrect results if they're being called by multiple threads at the same time using the same value for the name, I believe the owner must be selected by the name in this case to avoid this

src/noit_filters_lmdb.c
```diff
@@ -1535,20 +1538,112 @@ noit_filters_lmdb_cull_unused() {
...
+  void *xml_data;
+  void (*xml_data_free)(void *);
+} lmdb_filter_set_closure_t;
+
+static void
+lmdb_set_filter_data_free(void *c) {
+  lmdb_filter_set_closure_t *lfsc = (lmdb_filter_set_closure_t *)c;
+  if (lfsc) {
+    lfsc->xml_data_free(lfsc->xml_data);
+    free(lfsc->error_string);
```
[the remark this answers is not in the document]
pamaddox: Fixed

src/noit_filters_lmdb.c
pamaddox: @pjulien - Not sure what you mean by "owner must be selected by the name". What do you suggest?

pjulien: ``eventer_set_owner`` selects a thread to run in the new pool.  I do not know much about this code but it seems it would be possible coming in from multiple threads that have the same value for the value ``name``

This could break the state management here

pjulien (approved):

src/noit_filters_lmdb.c
pamaddox: Temporarily fixed by setting the jobq concurrency down to 1 and disallowing multiple threads running filterset sets.

PR #818, opened 2022-11-15
Do Check Deletes In A Jobq

pamaddox:


src/noit_check_lmdb.c
```diff
@@ -1069,67 +1069,126 @@ noit_check_lmdb_remove_check_from_db(uuid_t checkid, mtev_boolean force) {
...
+} lmdb_delete_check_closure_t;
 
-  if(npats != 2) goto error;
-  if(mtev_uuid_parse(pats[1], checkid)) goto error;
+static void
+noit_check_lmdb_delete_check_free_closure(void *closure) {
+  lmdb_delete_check_closure_t *ldcc = (lmdb_delete_check_closure_t *)closure;
+  if (ldcc) {
+    free(ldcc);
+  }
```
mcieslinski, line 1083: I feel confused by this. First, I didn't think a C cast could fail like this, and second doesn't `free` take a `void *` anyway?

mcieslinski: I guess the thing that threw me off, then, is that, at least according to the docs, `free` can safely take `NULL` as a parameter. Doesn't matter to me either way, but it seemed odd.

mcieslinski (approved):
Seems ok

src/noit_check_lmdb.c
pamaddox: An example of how this typically works can be found here:

https://github.com/circonus-labs/reconnoiter/blob/master/src/noit_check_lmdb.c#L618

Here, the struct has internal members requiring manual free. Hence, the need to cast. Despite the fact that there are no internal members that need freeing at the moment, if we add some at a later date, we will need the cast. The cast is extremely cheap and documents the type, since mtev eventer callbacks take a void*. This cast lets anyone trying to modify the code what the actual type is.

pjulien: ``free`` does accept null, the ``if`` statement was not required.  I usually don't like adding code for nothing but in this case it's very likely that more fields could be added in the future.  Can call ``free`` on it but can derefence any of its future members

src/noit_check_lmdb.c
```diff
@@ -1069,67 +1069,126 @@ noit_check_lmdb_remove_check_from_db(uuid_t checkid, mtev_boolean force) {
...
-  if(check) {
-    if(!noit_poller_deschedule(check->checkid, mtev_true, mtev_false)) {
-      just_mark = mtev_true;
+static int
+noit_check_lmdb_delete_check_asynch(eventer_t e, int mask, void *closure,
+                                    struct timeval *now) {
+  lmdb_delete_check_closure_t *ldcc = (lmdb_delete_check_closure_t *)closure;
+  mtev_log_go_synch();
+  if(mask == EVENTER_ASYNCH_WORK) {
+    mtevL(mtev_error, "PHIL: ASYNCH 1\n");
```
pjulien, line 1107: please remove this line

src/noitd.c
```diff
@@ -330,6 +330,7 @@ noitd_jobqs_init(void) {
    * TODO: Fan out filters across multiple threads by name
    * to run in parallel */
   mtev_main_eventer_config("jobq_lmdb_set_filter", "1,1,1,gc");
+  mtev_main_eventer_config("jobq_lmdb_check_delete", "1,1,1,gc");
```
pjulien, line 333: comment please on why it needs to be 1

PR #824, opened 2022-11-22
Improved Filterset/Check LMDB Set Error Logging

pamaddox:
Add extra logging for failures to fetch checks/filtersets immediately after loading them.

### DIFF
--- a/src/noit_check_lmdb.c
+++ b/src/noit_check_lmdb.c
@@ -39,6 +39,11 @@
 
 mtev_boolean lmdb_checks_support_inheritence = mtev_false;
 
+typedef enum {
+  CHECK_SHOW_SOURCE_SHOW = 0,
+  CHECK_SHOW_SOURCE_SET
+} noit_check_lmdb_show_source_e;
+
 static int noit_check_lmdb_show_check_json(mtev_http_rest_closure_t *restc, 
                                            uuid_t checkid) {
   return rest_show_check_json(restc, checkid);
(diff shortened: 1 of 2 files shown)


src/noit_check_lmdb.c
```diff
@@ -307,7 +312,8 @@ int noit_check_lmdb_show_checks(mtev_http_rest_closure_t *restc, int npats, char
   return 0;
 }
 
-int noit_check_lmdb_show_check(mtev_http_rest_closure_t *restc, int npats, char **pats) {
+int __noit_check_lmdb_show_check_internal(mtev_http_rest_closure_t *restc, int npats,
```
pjulien, line 315: This is not kosher, any symbol in the global scope that starts with `_` is reserved for the compiler and library implementation per ANSI C

If you want to communicate that this is private, please use ``static`` or better yet, move to C++ and put this function as a private method of a class or in a private namespace

src/noit_check_lmdb.c
```diff
@@ -307,7 +312,8 @@ int noit_check_lmdb_show_checks(mtev_http_rest_closure_t *restc, int npats, char
...
+                                          char **pats, noit_check_lmdb_show_source_e source) {
```
pjulien, line 316: ``source`` should be const

src/noit_filters_lmdb.c
```diff
@@ -1368,9 +1375,10 @@ noit_filters_lmdb_rest_show_filters(mtev_http_rest_closure_t *restc,
   return 0;
 }
 
-int
-noit_filters_lmdb_rest_show_filter(mtev_http_rest_closure_t *restc,
-                                   int npats, char **pats) {
+static int
+__noit_filters_lmdb_rest_show_filter_internal(mtev_http_rest_closure_t *restc,
```
pjulien, line 1379: same

src/noit_filters_lmdb.c
```diff
@@ -79,6 +79,11 @@ typedef enum {
   FILTERSET_RULE_METRIC_TYPE
 } noit_filter_lmdb_rule_type_e;
 
+typedef enum {
+  FILTERSET_SHOW_SOURCE_SHOW = 0,
+  FILTERSET_SHOW_SOURCE_SET
+} noit_filter_lmdb_show_source_e;
```
pjulien, line 85: could easily be the same enum reused for both use cases

PR #825, opened 2022-11-29
CIRC-9456 - Allow creation of implicit "__name" tags

RobBoeckermann:
tags created using `noit_metric_tagset_builder_add_one` are capped at 256 characters. 
this PR adds a new endpoint, `noit_metric_tagset_builder_add_one_implicit` to allow for creation of `__name:` tags which are capped at the max name length of 4096 characters (4095 in practice) + 7 characters (the length of the "__name:" key).

merging this functionality would add official support for implicit `__name` tags while leaving the existing `noit_metric_tagset_builder_add_one` functionality unchanged. maintaining the existing 256 character limit for all tags (aside from `__name`) is important for protecting against resource starvation.


src/noit_message_decoder.c
```diff
@@ -413,9 +413,10 @@ noit_metric_tags_count(const char *str, size_t strlen) {
   return rval;
 }
 
-const char *
-noit_metric_tags_parse_one(const char *tagnm, size_t tagnmlen,
-                           noit_metric_tag_t *output, mtev_boolean *toolong) {
+const char *noit_metric_tags_parse_one(const char *tagnm, size_t tagnmlen,
+                                       noit_metric_tag_t *output,
+                                       mtev_boolean *toolong,
+                                       int max_tag_length) {
```
pjulien, line 419:
```suggestion
                                       const int32_t max_tag_length) {
```

src/noit_message_decoder.c
```diff
@@ -413,9 +413,10 @@ noit_metric_tags_count(const char *str, size_t strlen) {
   return rval;
 }
 
-const char *
-noit_metric_tags_parse_one(const char *tagnm, size_t tagnmlen,
-                           noit_metric_tag_t *output, mtev_boolean *toolong) {
+const char *noit_metric_tags_parse_one(const char *tagnm, size_t tagnmlen,
```
pjulien, line 416:
```suggestion
const char *noit_metric_tags_parse_one(const char *const tagnm, const size_t tagnmlen,
```

src/noit_metric.h
```diff
@@ -101,6 +101,7 @@ typedef enum {
 } noit_message_type;
 
 #define NOIT_TAG_MAX_PAIR_LEN 256
+#define NOIT_NAME_MAX_PAIR_LEN MAX_METRIC_TAGGED_NAME + sizeof("__name:")
```
pjulien, line 104: - This is part of the public API and cannot change, i.e., you cannot change the value of this constant.  
- Beyond the API, and the ABI, this doesn't work either, because other than this one exception, we don't allow anything over 256 bytes anyway!  This constant must still reflect this otherwise we're allowing all tags to reach 4k

test/test_tags.c
```diff
@@ -726,7 +726,7 @@ void test_tag_at_limit(void) {
 
   assert(tag_name_len - 1 == NOIT_TAG_MAX_PAIR_LEN);
 
-  noit_metric_tags_parse_one(tag_name, tag_name_len - 1, &tag, &too_long);
+  noit_metric_tags_parse_one(tag_name, tag_name_len - 1, &tag, &too_long, NOIT_TAG_MAX_PAIR_LEN);
```
pjulien, line 729: Explicitly asked for a new function so that we wouldn't need to change the call sites... Furthermore, unless this tag is ``__name``, this is allowing tags that are way too big.  Same for the other call sites

src/noit_message_decoder.c
```diff
@@ -461,7 +462,10 @@ noit_metric_tags_parse_one(const char *tagnm, size_t tagnmlen,
 static int
 tag_canonical_size(noit_metric_tag_t *tag) {
   int len;
-  char dbuff[NOIT_TAG_MAX_PAIR_LEN];
+  // TODO: array increased from 256 to 4096+7 to allow for longer tag names.
```
pjulien, line 465: Memory wise it does not matter because it's on the stack and that's pretty much free.  However, nowhere does this check if ``max_tag_length`` is an acceptable value.  The use of ``mtev_dyn_buffer_t`` would be appropriate in this circumstances

src/noit_metric.h
RobBoeckermann: this is a new definition. not a change to the old value

test/test_tags.c
RobBoeckermann: this function is not public and is called by `noit_metric_tagset_builder_add_one` and `noit_metric_tagset_builder_add_name`

RobBoeckermann: I can rename it to make it clear that it is not exposed

src/noit_message_decoder.c
RobBoeckermann: is `int32_t` to ensure that this behaves correctly on both 32 and 64 bit machines? are there any other reasons to use `int32_t` over `int`?
should I make this `uint32_t` while I'm at it since it should never be negative?

src/noit_metric.h
pjulien: Please use ``inline const`` for new stuff then

test/test_tags.c
pjulien: No, that's fine, as long the bounds are properly checked

src/noit_message_decoder.c
pjulien: > is `int32_t` to ensure that this behaves correctly on both 32 and 64 bit machines? are there any other reasons to use `int32_t` over `int`? 

https://www.google.com/search?q=why+prefer+int32_t+in+c&oq=why+prefer+int32_t+&aqs=chrome.1.69i57j33i160l4.8791j0j7&sourceid=chrome&ie=UTF-8

> should I make this `uint32_t` while I'm at it since it should never be negative?

Prefer signed to unsigned in C unless you're sure you know all the problems that come with ``unsigned`` are... with the exception of ``site_t`` or any other non-arithmetic use.

test/test_tags.c
RobBoeckermann: @pjulien `noit_metric_tags_parse_one` is in `noit_metric.h`. Why are you no longer concerned about changing the call site? There are no public repositories on github that call this function (aside from reconnoiter forks), but the function is still exposed. In updating this, are we assuming that no one else is calling this function?

src/noit_message_decoder.c
RobBoeckermann: > The use of `mtev_dyn_buffer_t` would be appropriate in this circumstances

I will replace
```
char dbuff[NOIT_TAG_MAX_PAIR_LEN];
```
with `mtev_dyn_buffer_t`

> nowhere does this check if `max_tag_length` is an acceptable value. 

assuming this comment refers to `noit_metric_tags_parse_one`, would it be safer to implement the `max_tag_length` parameter as an enum so that it can only be 256 or 4096?

test/test_tags.c
pjulien: because you said it's private, hence it's private

src/noit_message_decoder.c
pjulien: No because someone else might need a value that's different than that and this is a private function

test/test_tags.c
RobBoeckermann: I was wrong about that; this is a public function (it is declared in the header).
I will try to figure out a way to implement this without changing the call to `noit_metric_tags_parse_one`

pjulien: Make a new function please, call the new function, with extra arguments, from the old functions

src/noit_message_decoder.c
RobBoeckermann: > nowhere does this check if `max_tag_length` is an acceptable value.

if a `max_tag_length` > 4096 is passed to the function, should we return as a failure or manually reduce the `max_tag_length` to 4096?

pjulien: That doesn't matter, what matters is the array being large enough

```c
char dbuff[NOIT_TAG_MAX_PAIR_LEN];
```
in combination with a value higher than 4k would corrupt the stack but you're using ``mtev_dyn_buffer`` now

RobBoeckermann: Ah, I understand. Thanks!

RobBoeckermann: @pjulien I can not pass a `mtev_dyn_buffer_t` to `noit_metric_tagset_decode_tag` because they would mutate the buffer without ensuring that the buffer is big enough to handle the new data.
using `mtev_dyn_buffer_t` would require writing a new function to handle the functionality of `noit_metric_tagset_decode_tag` for a `mtev_dyn_buffer_t`. The same would need to be done for all functions that work with the other arrays in the code that are currently sized for the max tag length (256) and need to also be updated to support `__name`.
This isn't very practical, so I think we should, instead, perform a check to make sure that the requested `max_tag_length` is less than the size of the array (4096) and also increase the array sizes to 65535 as @pamaddox pointed out is already being done within `check_metadata_alter_specific`:
```
  char decoded[NOIT_TAG_MAX_PAIR_LEN + 1];
  size_t decoded_len =
    noit_metric_tagset_decode_tag(decoded, sizeof(decoded) - 1, tag, strlen(tag));
```
https://circonus.slack.com/archives/C0J5RAKDJ/p1669845638697289

pjulien: Use ``mtev_dyn_buffer_ensure``

src/noit_message_decoder.c
```diff
@@ -458,15 +459,29 @@ noit_metric_tags_parse_one(const char *tagnm, size_t tagnmlen,
...
-  int len;
-  char dbuff[NOIT_TAG_MAX_PAIR_LEN];
-  len = noit_metric_tagset_decode_tag(dbuff, sizeof(dbuff), tag->tag, tag->total_size);
-  if(len < 0) return 0;
-  len = noit_metric_tagset_encode_tag(dbuff, sizeof(dbuff), dbuff, len);
-  if(len < 0) return 0;
-  return len;
+  mtev_dyn_buffer_t dbuff;
+  mtev_dyn_buffer_init(&dbuff);
+  size_t max_decode_len = mtev_b64_max_decode_len(tag->total_size);
```
pjulien, line 473:
```suggestion
  const size_t max_decode_len = mtev_b64_max_decode_len(tag->total_size);
```

src/noit_message_decoder.c
```diff
@@ -458,15 +459,29 @@ noit_metric_tags_parse_one(const char *tagnm, size_t tagnmlen,
...
+  mtev_dyn_buffer_ensure(&dbuff, mtev_b64_encode_len(max_decode_len));
+  int len = noit_metric_tagset_decode_tag((char *)mtev_dyn_buffer_data(&dbuff), max_decode_len, tag->tag, tag->total_size);
+  if(len >= 0){
```
pjulien, line 476: clang-format your new/updated code please

src/noit_message_decoder.c
```diff
@@ -458,15 +459,29 @@ noit_metric_tags_parse_one(const char *tagnm, size_t tagnmlen,
...
-  char dbuff[NOIT_TAG_MAX_PAIR_LEN];
-  len = noit_metric_tagset_decode_tag(dbuff, sizeof(dbuff), tag->tag, tag->total_size);
-  if(len < 0) return 0;
-  len = noit_metric_tagset_encode_tag(dbuff, sizeof(dbuff), dbuff, len);
-  if(len < 0) return 0;
-  return len;
+  mtev_dyn_buffer_t dbuff;
+  mtev_dyn_buffer_init(&dbuff);
+  size_t max_decode_len = mtev_b64_max_decode_len(tag->total_size);
+  mtev_dyn_buffer_ensure(&dbuff, mtev_b64_encode_len(max_decode_len));
```
pjulien, line 474: I'm pretty sure this should be the decoded length

src/noit_message_decoder.c
```diff
@@ -728,12 +743,13 @@ noit_metric_tagset_builder_add_many(noit_metric_tagset_builder_t *builder,
 }
 
 mtev_boolean
-noit_metric_tagset_builder_add_one(noit_metric_tagset_builder_t *builder,
-                                   const char *tagstr, size_t tagstr_len) {
+add_tag_to_tagset_builder(noit_metric_tagset_builder_t *builder, const char *tagstr,
+                          size_t tagstr_len, const int32_t max_tag_length) {
```
pjulien, line 747:
```suggestion
                          const size_t tagstr_len, const int32_t max_tag_length) {
```

src/noit_message_decoder.c
```diff
@@ -749,6 +765,18 @@ noit_metric_tagset_builder_add_one(noit_metric_tagset_builder_t *builder,
...
 }
 
+mtev_boolean
+noit_metric_tagset_builder_add_one(noit_metric_tagset_builder_t *builder, const char *tagstr,
+                                    size_t tagstr_len) {
+  return add_tag_to_tagset_builder(builder, tagstr, tagstr_len, NOIT_TAG_MAX_PAIR_LEN);
+}
+
+mtev_boolean
+noit_metric_tagset_builder_add_one_name(noit_metric_tagset_builder_t *builder,
```
pjulien, line 775:
```suggestion
noit_metric_tagset_builder_add_one_implicit(noit_metric_tagset_builder_t *builder,
```

src/noit_message_decoder.c
```diff
@@ -749,6 +765,18 @@ noit_metric_tagset_builder_add_one(noit_metric_tagset_builder_t *builder,
...
+                                        const char *tagstr, size_t tagstr_len) {
+  return add_tag_to_tagset_builder(builder, tagstr, tagstr_len, MAX_METRIC_TAGGED_NAME + sizeof("__name:"));
```
pjulien, line 777:
```suggestion
  return add_tag_to_tagset_builder(builder, tagstr, tagstr_len, MAX_METRIC_TAGGED_NAME + sizeof("__name:") - 1);
```

src/noit_metric.c
```diff
@@ -349,16 +349,31 @@ noit_metric_tagset_encode_tag_ex(char *encoded_tag, size_t max_len,
                                  noit_metric_encode_type_t left,
                                  noit_metric_encode_type_t right)
 {
-  char scratch[NOIT_TAG_MAX_PAIR_LEN+1];
-  if(max_len > sizeof(scratch)) return -1;
+  mtev_dyn_buffer_t scratch;
+  mtev_dyn_buffer_init(&scratch);
+  mtev_dyn_buffer_ensure(&scratch, max_len);
+  if(max_len > mtev_dyn_buffer_size(&scratch)) {
+    //TODO: I am keeping this check only to confirm 'mtev_dyn_buffer_ensure' succeeded (since it
```
pjulien, line 356: It's fine, don't need this

src/noit_message_decoder.c
RobBoeckermann: I undid some formatting because it was inconsistent with the surrounding code. should I format the surrounding code to match or only format my updated lines?

src/noit_message_decoder.c
RobBoeckermann: for the call to 'noit_metric_tagset_decode_tag', yes, but then we would need to call 'mtev_dyn_buffer_ensure' again before calling 'noit_metric_tagset_encode_tag', so I just ensure the final size here so we only need to call 'mtev_dyn_buffer_ensure' once.

src/noit_message_decoder.c
RobBoeckermann: since we only intend to use this for names (all other implicit tags are short enough to be supported by the existing endpoint), I thought it would be safer to explicitly call this `name` so that no one tries to use it for something else.
do we anticipate new implicit tags that may use this in the future?

src/noit_metric.c
RobBoeckermann: we don't need the check or we don't need the comment?

src/noit_message_decoder.c
pjulien: No, ``implicit``, both Brian and I made the point it was to be ``implicit`` because we don't want to add anything if there are further needs

For the rationale, please re-read the original discussions on this matter

src/noit_metric.c
pjulien: both

src/noit_message_decoder.c
pjulien: this:

> clang-format your new/updated code please

pjulien (commented):
I don't remember where we are with this... can I get an up to date status please

pjulien (approved):

RobBoeckermann:
@pjulien This PR was on hold while I was implementing the implicit tags within snowth (in case anything I discovered through implementing it changed the approach to the design). I'm confident now that this won't change, but am finalizing some cleanup within reconnoiter itself where we can use the new implicit functions instead of manually editing tagsets. I have not tested my last two commits yet.

pjulien (approved):

PR #833, opened 2023-01-19
CIRC-9456 - Implicit Tag Support

RobBoeckermann:
This PR adds the same implicit tag support as before:
1. Allow implicit tags with lengths up to 4103 (the maximum length of a `__name` tag pair).
2. Convert arrays to dynamic buffers so that they can be expanded to support the length of implicit tags when needed.
3. Allow filtering do be done on `__name` even if the max number of stream tags are already in used.

but has been reworked to:
4. Allow adding multiple implicit tags to the same tagset.
5. Allow adding multiple tags at the same time.

`noit_metric_add_implicit_tags_to_tagset` can be passed a *char containing a comma separated list of implicit tags and can now optionally be used to get the canonical name as well.

I refactored to keep code as DRY as possible by only maintaining one function, `add_tags_to_tagset_builder`, that can be used to add "one" or "many" tags to a tagset (instead of having two different functions for this logic).
The `add_one` and `add many` endpoints have the same behavior, but are kept as separate functions to prevent altering public function signatures.


src/noit_metric_tag_search.c
```diff
@@ -1373,8 +1354,27 @@ noit_metric_tag_search_evaluate_against_metric_id(const noit_metric_tag_search_a
...
+  char implicit_str[NOIT_IMPLICIT_TAG_MAX_PAIR_LEN + 1 +
+                    sizeof("__check_uuid:") + UUID_PRINTABLE_STRING_LENGTH];
+  size_t uuid_index =
+      snprintf(implicit_str, sizeof(implicit_str),
+               "__name:%.*s,__check_uuid:", id->name_len, id->name);
+  mtev_uuid_unparse_lower(id->id, implicit_str + uuid_index);
+  noit_metric_add_implicit_tags_to_tagset(
+      implicit_str, uuid_index + UUID_PRINTABLE_STRING_LENGTH, &tagset_implicit,
+      NULL);
+  // TODO: should the following use NOIT_METRIC_TAGSET_STREAM or NOIT_METRIC_TAGSET_CHECK?
```
RobBoeckermann, line 1369: Does anyone know what `noit_metric_tagset_fixup_hook_invoke` does? Should it be passed `NOIT_METRIC_TAGSET_STREAM` or `NOIT_METRIC_TAGSET_CHECK` when being used for the implicit tag tagset containing `__name` and `__check_uuid`? 
`__name` was previously added to the stream tag tagset (using `NOIT_METRIC_TAGSET_STREAM`).
`__check_uuid` was previously added to the check tag tagset (using `NOIT_METRIC_TAGSET_CHECK`).
Do I need to separate these tags so that the fixup hook still uses those same values for `__name` and `__check_uuid`?

src/noit_filters.c
```diff
@@ -774,27 +775,29 @@ noit_apply_filterset(const char *filterset,
   const char *local_metric_name = noit_metric_get_full_metric_name(metric);
   mtev_gettimeofday(&now, NULL);
 
-  char encoded_nametag[NOIT_TAG_MAX_PAIR_LEN+1];
-  char decoded_nametag[NOIT_TAG_MAX_PAIR_LEN+1];
-  noit_metric_tag_t stags[MAX_TAGS], mtags[MAX_TAGS];
-  noit_metric_tagset_t stset = { .tags = stags, .tag_count = MAX_TAGS };
+  noit_metric_tag_t *stags = malloc((MAX_TAGS + 1) * sizeof(noit_metric_tag_t));
```
pjulien, line 778: use calloc

src/noit_filters.c
```diff
@@ -774,27 +775,29 @@ noit_apply_filterset(const char *filterset,
...
+  noit_metric_tag_t *mtags = malloc(MAX_TAGS * sizeof(noit_metric_tag_t));
```
pjulien, line 779: use calloc

src/noit_filters.c
```diff
@@ -774,27 +775,29 @@ noit_apply_filterset(const char *filterset,
...
-    stset.tag_count++;
-  }
+
+  /* create __name tagset*/
+  mtev_dyn_buffer_t dbuff;
+  mtev_dyn_buffer_init(&dbuff);
+  const size_t max_encode_len =
+      mtev_b64_encode_len(NOIT_IMPLICIT_TAG_MAX_PAIR_LEN);
+  mtev_dyn_buffer_ensure(&dbuff, max_encode_len);
+  char *nametag = (char *)mtev_dyn_buffer_data(&dbuff);
```
pjulien, line 794: This is silly, use ``mtev_dyn_buffer_add_printf`` instead

src/noit_filters.c
```diff
@@ -774,27 +775,29 @@ noit_apply_filterset(const char *filterset,
...
+  snprintf(nametag, max_encode_len, "__name%c%.*s", NOIT_TAG_DECODED_SEPARATOR,
+           mlen, local_metric_name);
+  size_t nlen = noit_metric_tagset_encode_tag(nametag, max_encode_len, nametag,
```
pjulien, line 797:
```suggestion
  const size_t nlen = noit_metric_tagset_encode_tag(nametag, max_encode_len, nametag,
```

src/noit_filters.c
```diff
@@ -774,27 +775,29 @@ noit_apply_filterset(const char *filterset,
...
+                                              strlen(nametag));
```
pjulien, line 798: The return value of ``snprintf`` is the length of the formatted string

src/noit_message_decoder.c
```diff
@@ -413,9 +414,9 @@ noit_metric_tags_count(const char *str, size_t strlen) {
   return rval;
 }
 
-const char *
-noit_metric_tags_parse_one(const char *tagnm, size_t tagnmlen,
-                           noit_metric_tag_t *output, mtev_boolean *toolong) {
+const char *parse_metric_tag(const char *const tagnm, const size_t tagnmlen,
+                             noit_metric_tag_t *output, mtev_boolean *toolong,
+                             const int32_t max_tag_length) {
```
pjulien, line 419:
```suggestion
                             const size_t max_tag_length) {
```

src/noit_metric.c
```diff
@@ -342,15 +342,10 @@ noit_metric_tagset_is_taggable_value(const char *val, size_t len)
...
-noit_metric_tagset_encode_tag_ex(char *encoded_tag, size_t max_len,
-                                 const char *decoded_tag, size_t decoded_len,
-                                 mtev_boolean for_search,
-                                 noit_metric_encode_type_t left,
-                                 noit_metric_encode_type_t right)
-{
-  char scratch[NOIT_TAG_MAX_PAIR_LEN+1];
-  if(max_len > sizeof(scratch)) return -1;
+static inline ssize_t noit_metric_tagset_encode_tag_ex(
+    char *encoded_tag, size_t max_len, const char *decoded_tag,
```
pjulien, line 346:
```suggestion
    char *encoded_tag, const size_t max_len, const char *decoded_tag,
```

src/noit_metric.c
```diff
@@ -342,15 +342,10 @@ noit_metric_tagset_is_taggable_value(const char *val, size_t len)
...
+    size_t decoded_len, mtev_boolean for_search, noit_metric_encode_type_t left,
```
pjulien, line 347:
```suggestion
    size_t decoded_len, mtev_boolean for_search, const noit_metric_encode_type_t left,
```

src/noit_metric.c
```diff
@@ -342,15 +342,10 @@ noit_metric_tagset_is_taggable_value(const char *val, size_t len)
...
+    noit_metric_encode_type_t right) {
```
pjulien, line 348:
```suggestion
    const noit_metric_encode_type_t right) {
```

src/noit_metric_tag_search.c
pjulien: > Does anyone know what noit_metric_tagset_fixup_hook_invoke does?

It does what all hooks do, it allows for an interception to alter the existing function

src/noit_metric_tag_search.c
```diff
@@ -1373,8 +1354,27 @@ noit_metric_tag_search_evaluate_against_metric_id(const noit_metric_tag_search_a
...
-  mtev_boolean ok = noit_metric_tag_search_evaluate_against_tags_multi(search, tagsets, 3);
+  // setup implicit tags
+  noit_metric_tagset_t tagset_implicit;
+  MKTAGSETCOPY(tagset_implicit);
+  char implicit_str[NOIT_IMPLICIT_TAG_MAX_PAIR_LEN + 1 +
+                    sizeof("__check_uuid:") + UUID_PRINTABLE_STRING_LENGTH];
+  size_t uuid_index =
+      snprintf(implicit_str, sizeof(implicit_str),
+               "__name:%.*s,__check_uuid:", id->name_len, id->name);
+  mtev_uuid_unparse_lower(id->id, implicit_str + uuid_index);
```
pjulien, line 1365: ``mtev_uuid_unparse_lower`` does not null terminate the output, is that ok here?

pjulien (commented):
- Please confirm you ran all the tests with ASAN
- It's also worth nothing that C++ has been added to ``libmtev``, this means the barrier to C++ has been removed from here as well

src/noit_message_decoder.c
RobBoeckermann: should 
```
static const int32_t NOIT_IMPLICIT_TAG_MAX_PAIR_LEN =
    MAX_METRIC_TAGGED_NAME + sizeof("__name:") - 1;
```
use `size_t` instead, as well? it is being passed to this parameter and is intended to always be used as a size.

pjulien: yes, ``sizeof`` returns a ``size_t``

src/noit_metric.c
RobBoeckermann: right is also mutable
`if(!for_search) left = right = '"';`

src/noit_metric_tag_search.c
RobBoeckermann: yes. `noit_metric_add_implicit_tags_to_tagset` uses a `size_t` to determine the end of the tag string. In this case, the size is the return of `snprintf` (the size of the formatted string) plus `UUID_PRINTABLE_STRING_LENGTH` (the size of the uuid appended by `mtev_uuid_unparse_lower`.

src/noit_message_decoder.c
```diff
@@ -611,28 +613,31 @@ ssize_t noit_metric_tags_canonical(const noit_metric_tag_t *tags,
...
+            (char *)mtev_dyn_buffer_data(&dbuff), max_ensured_len,
+            (char *)mtev_dyn_buffer_data(&dbuff), dlen);
       }
     }
     if (dlen < 0 || tagnmlen < dlen) {
       mtev_dyn_buffer_destroy(&dbuff);
       return -1;
     }
-    memcpy(tagnm, dbuff_data, dlen);
+    memcpy(tagnm, (char *)mtev_dyn_buffer_data(&dbuff), dlen);
```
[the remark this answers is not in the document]
RobBoeckermann: this line is copying the buffer's memory into the memory pointed to by `tagnm`. `mtev_dyn_buffer_add` is used for copying into the buffer's memory.

src/noit_filters.c
```diff
@@ -775,8 +775,8 @@ noit_apply_filterset(const char *filterset,
   const char *local_metric_name = noit_metric_get_full_metric_name(metric);
   mtev_gettimeofday(&now, NULL);
 
-  noit_metric_tag_t *stags = malloc((MAX_TAGS + 1) * sizeof(noit_metric_tag_t));
-  noit_metric_tag_t *mtags = malloc(MAX_TAGS * sizeof(noit_metric_tag_t));
+  noit_metric_tag_t *stags = calloc((MAX_TAGS + 1), sizeof(noit_metric_tag_t));
+  noit_metric_tag_t *mtags = calloc(MAX_TAGS, sizeof(noit_metric_tag_t));
```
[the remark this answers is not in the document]
RobBoeckermann: I don't know why I did this in the first place. we know the size at compile time, so we should just use 
```
noit_metric_tag_t stags[MAX_TAGS + 1], mtags[MAX_TAGS];
```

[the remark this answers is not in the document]
RobBoeckermann: Oh, I did this because, to be compatible with `noit_metric_tagset_builder_end` it needs to be able to `realloc`.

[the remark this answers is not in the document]
pjulien: >we can call noit_metric_tagset_init to properly set the tagset members to 0, but we still need to manually change the members afterwards.

Why is this not using a builder then?  Looking at the API, it's clear ``noit_metric_tagset_t`` is meant to be initialized from a builder and building tagets, like you're doing now, should use ``noit_metric_tagset_builder_t``

> tldr; there are a few functions at play that don't work very well together and I don't think we should prioritize refactor it right now.

This points to using the builder facility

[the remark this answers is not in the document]
RobBoeckermann: I agree that this should use the builder, but it's not immediately obvious how much refactoring is needed to update `noit_metric_parse_tags` to use the builder since this is not the only place `noit_metric_parse_tags` is used. 
If we want to refactor this, I think we should get the team's approval first since this is a little outside the scope of the ticket and I have other work lined up after implicit tags that may be higher priority.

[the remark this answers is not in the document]
pjulien: > I agree that this should use the builder, but it's not immediately obvious how much refactoring is needed to update

ok, nothing needed then, if the API is not meant to be used like this then that's fine

> a little outside the scope of the ticket

We would just make a ticket that blocks this one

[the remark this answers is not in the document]
RobBoeckermann: > We would just make a ticket that blocks this one
maybe related, but the builder functions already weren't being used here prior to the implicit tag changes, so this isn't anything that would block this ticket.

RobBoeckermann:
> * Please confirm you ran all the tests with ASAN

all tests are passing on an asan build. the only warnings are the leaks we are discussing that I believe belong to libmtev.

> * It's also worth nothing that C++ has been added to `libmtev`, this means the barrier to C++ has been removed from here as well

does this mean we can add c++ code in libmtev if/when we need it? is there a way for me to take advantage of this in this ticket?

pjulien:
> all tests are passing on an asan build. the only warnings are the leaks we are discussing that I believe belong to libmtev.

that does not look real to me

> 
> > * It's also worth nothing that C++ has been added to `libmtev`, this means the barrier to C++ has been removed from here as well
> 
> does this mean we can add c++ code in libmtev if/when we need it? is there a way for me to take advantage of this in this ticket?

Libmtev probably not, have to keep ABI and source compatibility, but it means it can be used from reconnoiter

test/test_tags.cpp
```diff
@@ -150,76 +148,123 @@ struct Matches testmatches[] = {
...
     {
-      { NULL, 0 }
+      { "and(tag:*too_long*)", mtev_false },
+      { NULL, mtev_false }
+    }
+  },
+  {
+    "tag1:value1,"
+    "tag2:value2,"
+    "tag3:this_is_the_max_length_of_a_tag_pair_000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
```
pjulien, line 220: I don't understand why these strings are not generated in code.  If the constant's value changes, these will fail.  The structure this belongs to can just be created on the heap and the string itself filled in code.

This was already mentioned before in comments

RobBoeckermann: whoops. I missed this one. I'll fix it 👍

src/noit_message_decoder.c
```diff
@@ -756,7 +756,10 @@ mtev_boolean add_tags_to_tagset_builder(noit_metric_tagset_builder_t *builder,
     } else if(next && (tag.total_size == tagstr_len || *next == ',')) {
       noit_metric_tagset_builder_el_t *el =
         (noit_metric_tagset_builder_el_t *) calloc(1, sizeof(noit_metric_tagset_builder_el_t));
-      el->tag = tag;
+      // copy tag to el->tag. may need to use the method from snowht to copy the
```
pjulien, line 759:
```suggestion
      // copy tag to el->tag. may need to use the method from snowth to copy the
```

but this comment doesn't help, we can't possibly know which method you're talking about.  It would help if you can tell me what that is here please

src/noit_message_decoder.c
```diff
@@ -756,7 +756,10 @@ mtev_boolean add_tags_to_tagset_builder(noit_metric_tagset_builder_t *builder,
...
+      // tag members individually
+      memcpy(&el->tag, &tag, sizeof(tag));
+      // el->tag = tag; // need to copy
```
pjulien, line 762:
```suggestion
```

src/noit_message_decoder.c
RobBoeckermann: This PR is still in a draft state (it just doesn't allow me to set it to a draft). I pushed these changes for Michael to test reproducing for me last week, but have since determined the issue and just need to trace it up the stack. I'll ping you when this is cleaned up and ready for review.

pjulien: That's fine but looking at both PRs, I would like to introduce APIs to do this properly and also consolidate how these things are allocated.  This code is very unsafe and anything that modifies it must make it safer first

RobBoeckermann: Sounds good. Let's see what changes are left after I find the actual cause of the bug. Some of these changes may have been unnecessary and go away.

src/noit_message_decoder.c
```diff
@@ -660,8 +701,8 @@ noit_metric_tagset_init_with_context(noit_metric_tagset_t *lookup,
     free((void *) tags);
     return -1;
   }
-  lookup->tags = realloc(tags, tag_count * sizeof(noit_metric_tag_t));
-  lookup->tag_count = tag_count;
+  lookup->tags = realloc(tags, parsed_tag_count * sizeof(noit_metric_tag_t));
```
RobBoeckermann, line 704: Prior to this change, this realloc was pointless because it would always realloc the same size that was previously calloced.
But, even with this change, what good is a realloc here if the tags have already been set by `noit_metric_tags_parse_with_context`? 
When would `parsed_tag_count` ever be different than `tag_count`? 
If `parsed_tag_count` > `tag_count`, `noit_metric_tags_parse_with_context` will have already written beyond the originally allocated memory. if that's the case, then this realloc has the potential to lose that memory if it can't extend the allocation and needs to copy elsewhere. `noit_metric_tags_parse_with_context` also runs the risk of overwriting other memory since it would be writing past the previously calloced memory.

src/noit_metric_tag_search.c
```diff
@@ -1332,34 +1332,15 @@ noit_metric_tag_search_evaluate_against_metric_id(const noit_metric_tag_search_a
...
-    mtev_memory_end();
-    return 0;
-  }
   MKTAGSETCOPY(tagset_check);
-  char uuid_str[13 + UUID_STR_LEN + 1];
-  strcpy(uuid_str, "__check_uuid:");
-  mtev_uuid_unparse_lower(id->id, uuid_str + 13);
-  noit_metric_tag_t uuid_tag = { .tag = uuid_str, .total_size = strlen(uuid_str), .category_size = 13 };
-  tagset_check.tags[tagset_check.tag_count++] = uuid_tag;
   if(noit_metric_tagset_fixup_hook_invoke(NOIT_METRIC_TAGSET_CHECK, &tagset_check) == MTEV_HOOK_ABORT) {
```
RobBoeckermann, line 1336: This invokes snowth's `check_tag_tagset_completer`, which will always return `MTEV_HOOK_CONTINUE`.
When `NOIT_METRIC_TAGSET_CHECK` is used, the function will check for the `__check_uuid` tag and log to `use_debug` if it's not found (all other options will just return immediately). 
In this case, it will never find the `__check_uuid` because it has been moved to `tagset_implicit` to allow `tagset_check` to utilize its max potential of `MAX_TAGS` tags instead of `MAX_TAGS -1` tags.
We use `NOIT_METRIC_TAGSET_CHECK` to correctly check for the `__check_uuid` tag when running this against the `tagset_implicit` tagset, but I am not sure what to do for this call for the `tagset_check` tagset.

The other options for `noit_metric_tagset_class_t` are `NOIT_METRIC_TAGSET_STREAM` and `NOIT_METRIC_TAGSET_MEASUREMENT`, which would be inappropriate to use here, even though they don't perform any logic in the function registered in snowth.

On the other hand, leaving this to use `NOIT_METRIC_TAGSET_CHECK` doesn't feel right either, because we are unnecessarily checking for the existence of the `__check_uuid` tag that we know will never be in the tagset (and logging the "missing" tag to `use_debug' every time).

We could maybe just not invoke this function for `tagset_check`, but I don't know how other projects may be registering this, so I don't want to assume other implementations work the same as snowth.


The cleanest solution is to create a new enum called `NOIT_METRIC_TAGSET_IMPLICIT` and change snowth's `check_tag_tagset_completer` to only perform the `__check_uuid` logic for the new enum (instead of the CHECK enum that it currently performs it for), but introducing a new enum to reconnoiter (and changing the existing logic surrounding them) may break other implementations of `noit_metric_tagset_fixup_hook_invoke`.
I have yet to check if we have any other implementation of this, but the project is open-source, so I'd like to discuss any of these changes before I make them on my own, even if snowth is circonus's only implementation of this function.

There's also the timing issue of deploying these changes to both reconnoiter and snowth in a way that does not break the functionality (maybe perform the `__check_uuid` logic on both CHECK and IMPLICIT enums (and sacrifice the extra processing and logs) until we are sure all libnoits have been updated and are comfortable updating snowth to only perform the logic on the IMPLICIT enum.

test/test_tags.cpp
```diff
@@ -109,18 +110,27 @@ const char *testtags[][2] = {
...
   struct {
     const char *query;
     mtev_boolean match;
     uint64_t bench_ns;
   } queries[14];
 };
 
+std::string max_length_tag_pair = "max_length_tag:max_length" + std::string(NOIT_TAG_MAX_PAIR_LEN - (sizeof("max_length_tag:max_length") - 1) - 3, '0') + "256";
+std::string too_long_tag_pair = "too_long_tag:too_long" + std::string((NOIT_TAG_MAX_PAIR_LEN + 1) - (sizeof("too_long_tag:too_long") - 1) - 3, '0') + "257";
+std::string max_length_test(std::string("tag1:value1,") +
```
pjulien, line 122: use ``fmt::format`` to build your strings please

test/test_tags.cpp
```diff
@@ -150,76 +159,110 @@ struct Matches testmatches[] = {
...
+      { "and(too_long_tag:too_long*)", mtev_false },
+      { "and(tag4:value4)", mtev_true },
+      { NULL, mtev_false }
+    }
+  }
+};
+
+std::string max_length_implicit_tag_pair = "__name:max_length" + std::string(NOIT_IMPLICIT_TAG_MAX_PAIR_LEN - (sizeof("__name:max_length") - 1), '0');
+std::string too_long_implicit_tag_pair = "__name:too_long" + std::string((NOIT_IMPLICIT_TAG_MAX_PAIR_LEN + 1) - (sizeof("__name:too_long") - 1), '0');
+std::string check_uuid_tag_pair = "__check_uuid:b946274b-183d-4553-814a-ada8130c560d";
```
pjulien, line 252: same, use ``fmt::format``

test/test_tags.cpp
```diff
@@ -278,7 +321,7 @@ void test_tag_decode()
 void test_ast_decode()
 {
   int erroroffset;
-  noit_metric_tag_search_ast_t *ast, *not;
+  noit_metric_tag_search_ast_t *ast, *_not;
```
pjulien, line 324: limit the underscore prefix for fields.

test/test_tags.cpp
```diff
@@ -719,11 +810,11 @@ void query_argument_swapping(void) {
   }
 }
 
-void loop(char *str) {
+void loop(std::string str) {
```
pjulien, line 813:
```suggestion
void loop(const std::string &str) {
```

test/test_tags.cpp
```diff
@@ -486,7 +528,59 @@ void test_tag_match()
...
+  test_assert(max_length_implicit_tag_pair.length() == NOIT_IMPLICIT_TAG_MAX_PAIR_LEN);
+  test_assert(too_long_implicit_tag_pair.length() > NOIT_IMPLICIT_TAG_MAX_PAIR_LEN);
+  test_assert(check_uuid_tag_pair.length() == sizeof("__check_uuid:") - 1 + UUID_STR_LEN);
+
+  for (size_t i = 0;
+       i < sizeof(implicit_testmatches) / sizeof(*implicit_testmatches); i++) {
+    char *canonical = NULL;
+    test_assert_namef(true, "testing tagset '%s'",
+                      implicit_testmatches[i].tagstring);
+    memset(&tagset, 0, sizeof(tagset));
```
pjulien, line 551:
```suggestion
    tagset = {};
```

src/noit_message_decoder.c
pjulien: don't know, would need to pretty much take over this ticket to figure that out

src/noit_metric_tag_search.c
pjulien: > This invokes snowth's check_tag_tagset_completer, which will always return MTEV_HOOK_CONTINUE.
When NOIT_METRIC_TAGSET_CHECK is used, the function will check for the __check_uuid tag and log to use_debug if it's not found (all other options will just return immediately).
In this case, it will never find the __check_uuid because it has been moved to tagset_implicit to allow tagset_check to utilize its max potential of MAX_TAGS tags instead of MAX_TAGS -1 tags.
We use NOIT_METRIC_TAGSET_CHECK to correctly check for the __check_uuid tag when running this against the tagset_implicit tagset, but I am not sure what to do for this call for the tagset_check tagset.

It seems to be also null terminating all the uuid strings.  This behavior should be retained and verified

> The cleanest solution is to create a new enum called NOIT_METRIC_TAGSET_IMPLICIT and change snowth's check_tag_tagset_completer to only perform the __check_uuid logic for the new enum (instead of the CHECK enum that it currently performs it for), but introducing a new enum to reconnoiter (and changing the existing logic surrounding them) may break other implementations of noit_metric_tagset_fixup_hook_invoke.
I have yet to check if we have any other implementation of this, but the project is open-source, so I'd like to discuss any of these changes before I make them on my own, even if snowth is circonus's only implementation of this function.

don't understand what that means

src/noit_metric_tag_search.c
```diff
@@ -1373,8 +1354,28 @@ noit_metric_tag_search_evaluate_against_metric_id(const noit_metric_tag_search_a
...
   }
 
-  const noit_metric_tagset_t *tagsets[3] = { &tagset_check, &tagset_stream, &tagset_measurement };
-  mtev_boolean ok = noit_metric_tag_search_evaluate_against_tags_multi(search, tagsets, 3);
+  // setup implicit tags
+  noit_metric_tagset_t tagset_implicit = {};
+  noit_metric_tagset_init(&tagset_implicit, 0, 0);
+  char implicit_str[NOIT_IMPLICIT_TAG_MAX_PAIR_LEN + 1 +
+                    sizeof("__check_uuid:") + UUID_PRINTABLE_STRING_LENGTH];
+  size_t uuid_index =
```
pjulien, line 1362:
```suggestion
  const size_t uuid_index =
```

src/noit_message_decoder.c
RobBoeckermann: this realloc is correct in it's current (`parsed_tag_count`) state.

`noit_metric_tags_count` returns the number of commas + 1 as the `tag_count`.
`parsed_tag_count` could end up being returned less than `tag_count` if tags do not match the validation function or if there are duplicate tags.
This means that the realloc will never allocate __more__ memory. It can only ever reduce the allocated memory.

src/noit_metric_tag_search.c
RobBoeckermann: There was no need to change anything outside of this function. I changed this back to adding the implicit tags directly to the tagsets already being created in this scope, but made changes to allow this to work even if the MAX_TAGS are already being used. Since this is copying the tags to a local tagset, we can just make that local tagset size MAX_TAGS+1 (there are no restrictions within the eval function called on these tagsets that would actually restrict the tag count to MAX_TAGS).

pjulien (approved):

RobBoeckermann:
these reconnoiter changes are breaking the picker tests on master:
```
[ RUN      ] matching/basic_matching/basic_matching_spec.lua @ 79: correctly filter input based on rules verifies that we matched the numeric ruleset
/opt/circonus/libexec/mtev/lua/mtev/HttpClient.lua:167: no response

stack traceback:
        /opt/circonus/libexec/mtev/lua/mtev/HttpClient.lua:167: in function 'get_headers'
        /opt/circonus/libexec/mtev/lua/mtev/HttpClient.lua:324: in function 'get_response'
        /opt/circonus/libexec/mtev/lua/mtevbusted/api.lua:129: in function 'load_data'
        matching/basic_matching/basic_matching_spec.lua:80: in function <matching/basic_matching/basic_matching_spec.lua:79>

[  ERROR   ] matching/basic_matching/basic_matching_spec.lua @ 79: correctly filter input based on rules verifies that we matched the numeric ruleset (29.53 ms)
```

This won't be merged until this is resolved.

RobBoeckermann:
@pjulien there is a bug in a picker test when using these changes to libnoit:
```
matching/basic_matching/basic_matching_spec.lua @ 80: correctly filter input based on rules verifies that we matched the numeric ruleset
matching/basic_matching/basic_matching_spec.lua:86: Expected objects to be equal.
Passed in:
(number) 0
Expected:
(number) 1
```

This is failing because the `debug/matching` log:
```[2023-06-20 20:13:56.996525] [debug/matching] [t@26651/e:default/2,integrations/selector/selector_integration1: b52856c0-6eb6-44e6-e016-c2794bae6b73:a_numeric_metric forwarded to numeric_ruleset_test_exchange:numeric_ruleset_test_route```
is not found in the log file.

This log is written by the picker's `selector_integration.cpp:168`, but this code is not being hit because, higher in the callstack, a while loop on picker's `metrics.cpp:156` is not being entered due to `noit_metric_director_lane_next` is returning `nullptr`.

This is `nullptr` because the metric director's fifo is empty, so `ck_fifo_spsc_dequeue` is unable to return a pointer.

I'm not sure yet why the fifo is not being added to, but am continuing to look into it.

pjulien (commented):
can you run try to run this under lsan/asan please?

PR #840, opened 2023-03-16
Small Performance Improvements

pamaddox:
* Use `mtev_hash_size` instead of looping through every value to count when making a metric bundle.
* Optimistic insertion in the httptrap module - rather than retrieving a value to determine if it's there already, try to insert a new value and deduce the existence of a conflict from insert failure.

### DIFF
--- a/src/noit_check_log.c
+++ b/src/noit_check_log.c
@@ -656,7 +656,6 @@ noit_check_log_bundle_serialize(mtev_log_stream_t ls, noit_check_t *check, const
   int rv = 0;
   static char *ip_str = "ip";
   char uuid_str[256*3+37];
-  mtev_hash_iter iter = MTEV_HASH_ITER_ZERO;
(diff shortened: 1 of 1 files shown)


pjulien (commented):
can you move ``iter`` please to the closest enclosing scope.  I'm worried they will be reused incorrectly post some future modification

src/noit_check_log.c
```diff
@@ -658,7 +657,6 @@ noit_check_log_bundle_serialize(mtev_log_stream_t ls, noit_check_t *check, const
   static char *ip_str = "ip";
   char uuid_str[256*3+37];
   mtev_hash_iter iter = MTEV_HASH_ITER_ZERO;
```
pjulien, line 659: This looks like it can be moved to inside the ``if`` on line 735

pjulien (approved):

PR #843, opened 2023-03-20
Use Explicit Else When Finishing With httptrap Payload

pamaddox:


bdunavant (approved):

src/modules/httptrap.c
```diff
@@ -802,8 +802,7 @@ rest_get_json_upload(mtev_http_rest_closure_t *restc,
       rxc->complete = 1;
       _YD("no more data, finishing YAJL parse\n");
       yajl_complete_parse(rxc->parser);
-    }
-    if (++loop_count % 25 == 0 && !rxc->complete) {
+    } else if (++loop_count % 25 == 0 && !rxc->complete) {
```
pjulien, line 805: This is not equivalent to what was there before.  Before ``++loop_count`` regardless and that's no longer the case

pjulien: > Does it matter?

You tell me, I don't know this code but I know before and after are not equivalent

pamaddox: I do not think it matters, but if it's being commented on, I assume you are seeing something I am not.

bdunavant: That loop_count is new, and is not needed if the previous `if` was true.

PR #847, opened 2023-03-23
Revert httptrap ACO

pamaddox:
We don't strictly need this and there's performance/fanout issues associated with this. As it's not a large performance improvement, just remove this for now.


pjulien (approved):

mcieslinski (approved):
Seems ok

PR #850, opened 2023-05-23
CIRC-10318 - add lua debugging

RobBoeckermann:
added and improved lua debugging.

I plan to update snowth with these changes as well:
* no maintenance required if the path to the debugger changes due to a version bump
* you only need to add `utils.enableLuaDebugging()` after the requires in order to enable debugging of a lua file.
  * I think it would be worth it to go through and add this line to every test, but that's up for discussion.
  * files still run normally if the user does not have the extension installed.
* add launch.json template
* add vscode extension recommendations


RobBoeckermann:
thoughts on enabling this for all tests by adding
```
utils.enableLuaDebugging()
```
to every lua test?

test/busted/lua-support/utils.lua
```diff
@@ -46,3 +47,34 @@ function postgres_reqs(bins)
...
+
+function getDebuggerPath()
+  local home_dir = os.getenv("HOME") or os.getenv("USERPROFILE")
+  local vscode_dir
+  if lfs.attributes(home_dir .. "/.vscode") then
+    vscode_dir = ".vscode"
+  elseif lfs.attributes(home_dir .. "/.vscode-server") then
+    vscode_dir = ".vscode-server"
+  else
+      return nil
```
pjulien, line 59: - indentation
- I guess I won't have a lua debugger :)

test/busted/lua-support/utils.lua
```diff
@@ -46,3 +47,34 @@ function postgres_reqs(bins)
   end
   return true
 end
+
+function getDebuggerPath()
+  local home_dir = os.getenv("HOME") or os.getenv("USERPROFILE")
+  local vscode_dir
```
pjulien, line 53:
```suggestion
  local ide_dir
```

RobBoeckermann: @pjulien This should only support the vscode extension. Do you have another process/tool that can utilize this?

vasuraman11: there's other ways besides vscode, I saw several lua debug engines including ones that work with the lightweight ZeroBrane lua ide, and there may also be a plug-in for gdb or something like that
So vscode is not the only way, it's just the one I picked because many of us can or do use vscode.
Just google what might work with your chosen toolset that you use.

RobBoeckermann:
> thoughts on enabling this for all tests by adding
> 
> ```
> utils.enableLuaDebugging()
> ```
> 
> to every lua test?

I'm sure there's shared code that's required by every test that I can call this from instead of adding it to the tests themselves.

test/busted/lua-support/utils.lua
RobBoeckermann: but this code is only to support this specific vscode extension. I don't think it's useful to any other IDEs, hence the "vscode" specific variable name.
I can change the name if we intend to modify this code to handle other IDEs/extensions in the future, but my understanding was that this was the only one in use right now.

vasuraman11: it's the only one in use right now so this is fine, but if/when someone wants to add support for other debuggers then we can figure it out then.
I would add it for every test, let the person who is debugging add it and commit it as part of their pr if they want.  Many of these tests might never be touched anytime soon so I would not waste your time...

vasuraman11 (approved):
LGTM we can add more later

test/busted/lua-support/utils.lua
RobBoeckermann: I found a way to enable it for all tests in snowth without needing to add it manually to each test: https://github.com/circonus/snowth/pull/4352
I plan to do the same for reconnoiter.

pjulien: > @pjulien This should only support the vscode extension. Do you have another process/tool that can utilize this?

I should be able to use it from Eclipse, maybe, but the point is you're selecting a debug path and support for additional IDEs should be possible

pjulien (approved):

PR #851, opened 2023-05-31
Adopt modern OpenSSL hash functions

esproul:


src/noit_metric_director.c
```diff
@@ -1114,7 +1115,13 @@ static mtev_boolean
 check_duplicate(char *payload, size_t payload_len) {
   if (dedupe) {
     unsigned char *digest = malloc(MD5_DIGEST_LENGTH);
-    MD5((unsigned char*)payload, payload_len, digest);
+    const EVP_MD *md = EVP_get_digestbyname("MD5");
```
pjulien, line 1118: should make our own md5 function in a single file that gets called from these sites instead of propagating ``OpenSSL`` code throughout

src/noit_metric_director.c
```diff
@@ -1232,6 +1239,8 @@ void noit_metric_director_init() {
     mtev_frrh_set_prob_function(search_miss_cache, generation_aware_displace);
   }
 
+  OpenSSL_add_all_digests();
```
pjulien, line 1242: should be done only once

src/noit_ssl10_compat.h
```diff
@@ -0,0 +1,31 @@
...
+}
+
+inline EVP_MD_CTX *EVP_MD_CTX_new(void)
+{
+  return OPENSSL_zalloc(sizeof(EVP_MD_CTX));
+}
+
+inline void EVP_MD_CTX_free(EVP_MD_CTX *ctx)
+{
+  EVP_MD_CTX_cleanup(ctx); OPENSSL_free(ctx);
```
pjulien, line 27: please use two lines, no reason to be on a single line

src/noit_ssl10_compat.h
```diff
@@ -0,0 +1,31 @@
...
+
+#include <string.h>
+#include <openssl/engine.h>
+
+inline void *OPENSSL_zalloc(size_t num)
+{
+   void *ret = OPENSSL_malloc(num);
+
+   if (ret != NULL)
+       memset(ret, 0, num);
```
pjulien, line 16:
```suggestion
       memset(ret, 0, num);
     }
```

src/noit_ssl10_compat.h
esproul: this is copied from libmtev's version of this header

src/noit_metric_director.c
esproul: There is both an md5 and a sha1 use case though, so we'd need two, and that's the same number of times we're currently using it, so I don't see what that gets us.

src/noit_metric_director.c
esproul: I agree, I just didn't see one global place to do it. We have 3 main contexts: noitd, stratcond, and libnoit. I couldn't find a common place to initialize this that would work for all. Additionally `test/test_tags.c` is a test-only binary that will eventually call into the `noit_metric.c` code but I don't want to have to add libcrypto to it just for that.

src/noit_ssl10_compat.h
esproul: FWIW both libmtev's version and this were just copied from OpenSSL's migration guide.

src/noit_metric.c
```diff
@@ -512,10 +514,16 @@ simple_sha1_hex(void *in, size_t len, char out[2 * SHA_DIGEST_LENGTH]) {
   unsigned char outb[SHA_DIGEST_LENGTH];
   static const char _hexchars[16] =
     {'0','1','2','3','4','5','6','7','8','9','a','b','c','d','e','f'};
-  SHA_CTX ctx;
-  SHA1_Init(&ctx);
-  SHA1_Update(&ctx, in, len);
-  SHA1_Final(outb, &ctx);
+
+  OpenSSL_add_all_digests();
```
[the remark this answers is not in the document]
esproul: See comment thread for noit_metric_director.c.

src/noit_metric_director.c
esproul: I think the most logical place to initialize digests would be in libnoit. Both noitd and stratcond build libnoit objects in, so all three contexts would be covered. However, there doesn't seem to be a good place to put this so I might need to add an object.

esproul: I think the most logical place to initialize digests would be in libnoit. Both noitd and stratcond build libnoit objects in, so all three contexts would be covered. However, there doesn't seem to be a good place to put this so I might need to add an object.

src/noit_metric.c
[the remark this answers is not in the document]
esproul: Removed in favor of a single call site.

src/noit_metric_director.c
esproul: Removed in favor of a single call site.

src/libnoit.h
```diff
@@ -0,0 +1,32 @@
...
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
```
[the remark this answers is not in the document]
esproul: Done.

pjulien (approved):

pamaddox (approved):

PR #856, opened 2023-07-03
Make grpc disabled and grpc ssl disabled the default

vasuraman11:
In testing of the broker with the enhanced otlphttp module, the default of grpc enabled and grpc ssl enabled causes a crash loop because grpc ssl is not configured properly (by the user).  In hindsight, we probably don't want port 4317 open by default, even if ssl were configured to some default.  And then we also don't want enabling grpc to result in immediate crash because ssl is enabled to try to protect the user from accidentally exposing unencrypted grpc port 4317 on the internet.
So...we just will make the default to disable grpc and disable grpc ssl, and let the user turn those on as they need/want them and on their own to configure them responsibly.


esproul (approved):
LGTM

PR #860, opened 2024-01-08
Fix Deadlock In Noit Check Resolver

pamaddox:
The noit_check_resolver_init() function was calling a function that could try to take a lock that the function already holds, leading to deadlock. Added a new, simple helper function to add the offending locking call to the event loop without calling it first, preventing the deadlock.

### DIFF
--- a/src/noit_check_resolver.c
+++ b/src/noit_check_resolver.c
@@ -532,6 +532,10 @@ int noit_check_resolver_loop(eventer_t e, int mask, void *c,
   return 0;
 }
 
+static inline void start_noit_check_resolver_loop() {
+  eventer_add_in_s_us(noit_check_resolver_loop, NULL, 1, 0);
+}
+
 static int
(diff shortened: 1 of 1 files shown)


src/noit_check_resolver.c
```diff
@@ -532,6 +532,10 @@ int noit_check_resolver_loop(eventer_t e, int mask, void *c,
   return 0;
 }
 
+static inline void start_noit_check_resolver_loop() {
+  eventer_add_in_s_us(noit_check_resolver_loop, NULL, 1, 0);
```
postwait, line 536: Why wait 1 second? Why not 0?

postwait (changes requested):

src/noit_check_resolver.c
pamaddox: Was copied from before. Really no reason for it not to be zero. I've changed it to 0.

postwait (approved):

PR #863, opened 2024-01-29
Fix Memory Leak In Marking Metrics Logged

pamaddox:
We were checking the return on whether we marked a metric logged, but the function being called duplicates the metric if it needs it, so we need to unconditionally clean up.


bdunavant (approved):
LGTM

PR #865, opened 2024-02-14
Allow Reading/Writing Batch Size Field In Noit Cluster

pamaddox:
Utilize mtev hooks to allow setting the `batch_size` field (and theoretically other fields if needed) from within noit_clustering. Allows the field to persist after restarts and to be saved when set via the API.


src/noit_clustering.c
```diff
@@ -76,11 +76,14 @@ MTEV_HOOK_IMPL(noit_should_run_check,
 
 #define MAX_CLUSTER_NODES 128 /* 128 this is insanely high */
 #define REPL_FAIL_WAIT_US 500000
+#define DEFAULT_BATCH_SIZE 500
```
bdunavant, line 79:
```suggestion
#define DEFAULT_BATCH_SIZE 10000
```
This default is too small at this point, it only needs to be small in cases like where people have 50MB snmp configs.

src/noit_clustering.c
```diff
@@ -898,6 +901,104 @@ cluster_topo_cb(void *closure,
...
+static mtev_hook_return_t
+reconnoiter_specific_read_node_config_cb(void *closure, mtev_cluster_t *cluster, uuid_t node_id, mtev_conf_section_t *conf) {
+  if(strcmp(mtev_cluster_get_name(cluster), NOIT_MTEV_CLUSTER_NAME)) {
+    return MTEV_HOOK_CONTINUE;
+  }
+  if (!conf) {
+    return MTEV_HOOK_CONTINUE;
+  }
+  // No node specific noit configs needed
+  return MTEV_HOOK_CONTINUE;
```
bdunavant, line 999: If it's always CONTINUE why are we doing these other things?

pamaddox: Documentation purposes. If we _did_ put any special code in there (where the comment is), we'd want to be able to continue or error out if there's a problem.

Technically, these node-specific hook callbacks are not necessary, as we are not using any node-specific stuff. I just have them there because I think it makes it clear where things would go later if we needed to add stuff; however, if you think it complicates/confuses things, I can take the functions completely out, as they do not need to be here.

bdunavant (approved):

PR #872, opened 2025-03-12
Set Metric Lengths Properly In Prometheus Handling

pamaddox:
Pass around the tagged and untagged metric lengths so we properly set the values on incoming kafka/prometheus data.


bdunavant (approved):
LGTM

PR #884, opened 2025-09-03
Add Functions To Canonicalize Prometheus Metric Names

pamaddox:
Added two new functions, `noit_prometheus_translate_snappy_data_canonical` and `noit_prometheus_metric_name_from_labels_canonical`, that will guarantee that the caller gets fully canonical metric names.


postwait (changes requested):
I believe this was intentional because callsites were already canonicalizing and given the cost we didn't want to force doing it twice.

I would add another function that promises to canonicalize... `_canonical`

postwait (approved):

